PR #26, opened 2018-08-07
Refactor plugin structure

tarokkk:

PR #28, opened 2018-08-08
Use crd name as config name instead of pattern to allow using wildcar…

baluchicken:
…d chars in pattern


baluchicken:
Closing this because this will be addressed in #26

PR #29, opened 2018-08-23
Refactor TLS and Monitoring

tarokkk:


tarokkk:
@baluchicken 

At this stage of the logging-operator we try to provide as failsafe configuration as possible. Since mTLS is a complex configuration and require deep knowledge of certificates etc we decided to postpone these kind of advanced features (like enable TLS only in 1 component). As we are proposing handling fluentbit and fluentd in CRDs (in usse: #30 ) I think we should return to this question when implementing that behaviour.

PR #116, opened 2019-07-31
Add splunk output and customize outputs plugins

wwojcik:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no|yes
| Deprecations?   | no

### Checklist
- [x] Implementation tested (with at least one cloud provider)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ADD . /go/src/github.com/banzaicloud/logging-operator
 RUN go install ./cmd/manager
 
 
-FROM alpine:3.8
+FROM alpine:3.9
 
 RUN apk add --no-cache ca-certificates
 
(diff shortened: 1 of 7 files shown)


pepov:
Hey @wwojcik thanks for your PR! 

Before moving on I would like to clarify one thing. We intentionally decided to avoid supporting custom fluent configuration snippets and would like to stick with this in the future as well. One reason for example is that we don't want to support things like templating or loading secrets in freeform configurations which would be necessary on the long run.
Because of this I would like to ask you to pull the custom output support out of this pull request. However feel free to open a new one with the custom output changes if you feel we should consider it and want discuss it further!

wwojcik:
ok I will do as you suggest

tarokkk:
For the V2 I opened a new issue for Splunk output: https://github.com/banzaicloud/logging-operator/issues/158
I think we will integrate it soon. But if you have time and port it to the V2 API that is welcomed as well!

PR #131, opened 2019-08-22
Add `path` argument to s3 plugin

worr:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes
| Related tickets | none

### What's in this PR?
This adds the ability to specify that path in s3 to send the logs. This preserves the existing path as the default.

### Why?
You may, in the case of buckets shared with other types of artifacts, want to change exactly where the logs get stored.

### Additional context
Testing:

* Built a docker image with my change
* redeployed logging operator
* Modified `s3-output` helm chart with the configuration
* Pushed a custom logger path

As far as edge cases, if there isn't a trailing `/` at the end of the path, the `path` will not be treated as a directory, but instead a prefix for the file. This appears to be a quirk of the fluentd s3 output plugin, not the operator itself.

If approved, I can make the changes in the upstream helm chart as well.

### Checklist

- [x] Implementation tested (with at least one cloud provider)

### DIFF
--- a/docs/plugins/s3.md
+++ b/docs/plugins/s3.md
@@ -18,6 +18,7 @@
 | bufferTimeWait | 10m |  |
 | timekey_use_utc | true |  |
 | format | json |  |
+| path | logs/${tag}/%Y/%m/%d/ |  |
 ## Plugin template
 ```
 <match {{ .pattern }}.** >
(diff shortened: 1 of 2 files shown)


pepov:
thanks @worr !

PR #134, opened 2019-09-02
Support TLS configuration with k8s.io/tls secrets

### LINKED ISSUE #133: support secrets of type kubernetes.io/tls
The operator doesn't support using TLS with secrets of type `kubernetes.io/tls`, since it generates a deployment and a configmap which assume paths like `serverCert`, while the standard path for this type of secret is `tls.crt`.

I would like either to be able to optionally generate deployments for fluentd and fluent-bit which project the standard items onto the paths, e.g.
```
      - name: fluent-tls
        secret:
          secretName: lala-fluentd-forward-tls
          items:
          - key: ca.crt
            path: caCert
          - key: tls.crt
            path: clientCert
          - key: tls.key
            path: clientKey
```
or alternatively a way to customize the fluentd and fluent-bit configuration templates to use the standard pathnames.

The alternative is to manually generate a generic secret in the way that the operator wants, however i am generating these certificates using cert-manager, which uses the standard `kubernetes.io/tls` type of secret which requires the particular paths to be `tls.key` and `tls.crt`.

kouk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #133 

### What's in this PR?
Modify the custom resources to optionally support specifying the certificate
and key for each service using a secret of type `kubernetes.io/tls`. This
requires that the CA certificate be passed in through a separate avenue, in
particular via a `ConfigMap` resource. The pre-existing method of configuration,
using a `Secret` that contained both cert/key pairs plus the CA cert remains
the default.

Also update the helm charts to optional support the new type of configuration.

Fixes: banzaicloud/logging-operator#133

### Why?

The `kubernetes.io/tls` type of secret is used in some notable products, such as [cert-manager](https://github.com/jetstack/cert-manager/) and is currently incompatible with `logging-operator`.

### Checklist

- [X] Implementation tested (with at least one cloud provider)
- [X] User guide and development docs updated (if needed)
- [X] Related Helm chart(s) updated (if needed)


charts/logging-operator-fluent/templates/fluentbit-cr.yaml
```diff
@@ -21,6 +22,15 @@ spec:
   {{- end }}
   tls:
     enabled: {{ .Values.tls.enabled }}
+{{- if $fluentbitUseGenericSecret }}
     secretName: {{ .Values.tls.secretName | default (include "logging-operator-fluent.fullname" .) }}
+    secretType: generic
+{{- else }}
+    secretName: {{ .Values.fluentd.tlsSecret }}
```
pepov:
```suggestion
    secretName: {{ .Values.fluentbit.tlsSecret }}
```

pepov:
hey @kouk doesn't certmanager include the ca in your tls certificates? I assumed we don't have to specify the ca cert separately, just use the one included in the tls secret.

kouk:
@pepov in the case of `certmanager` it does, you're right. Do you think we should, in the interest of simplicity, require the secret contain the CA certificate? `kubectl create secret` doesn't require a CA cert. However it's true that I'm not aware of a specific case where there would be a constraint to use a secret that did not contain the cert.

pkg/apis/logging/v1alpha1/fluentbit_types.go
```diff
@@ -44,6 +44,7 @@ type FluentbitSpec struct {
 type FluentbitTLS struct {
 	Enabled    bool   `json:"enabled"`
 	SecretName string `json:"secretName"`
+	SecretType string `json:"secretType"`
```
pepov: I suggest to let this be omitted for backwards compatibility
```suggestion
	SecretType string `json:"secretType,omitempty"`
```

pkg/apis/logging/v1alpha1/fluentd_types.go
```diff
@@ -48,6 +48,7 @@ type FluentdSpec struct {
 type FluentdTLS struct {
 	Enabled    bool   `json:"enabled"`
 	SecretName string `json:"secretName"`
+	SecretType string `json:"secretType"`
```
pepov: Let this be omitted for bc
```suggestion
	SecretType string `json:"secretType,omitempty"`
```

pkg/resources/fluentd/deployment.go
```diff
@@ -131,7 +131,7 @@ func generateVolumeMounts(fluentd *loggingv1alpha1.Fluentd) (v []corev1.VolumeMo
 	if fluentd.Spec.TLS.Enabled {
 		tlsRelatedVolume := []corev1.VolumeMount{
 			{
-				Name:      "fluentd-tls",
```
pepov: why is this necessary?

pkg/resources/fluentd/configmap.go
```diff
@@ -26,23 +26,38 @@ import (
...
+type fluentdCertConfig struct {
+	CertFile string
+	KeyFile  string
+}
+
+type fluentdTLSConfig struct {
+	Enabled    bool
+	SharedKey  string
+	CACertFile string
+	Cert       fluentdCertConfig
```
pepov: It's not necessary to split this into a separate struct I think

pkg/resources/fluentbit/configmap.go
```diff
@@ -24,14 +24,23 @@ import (
...
+type fluentbitCertConfig struct {
+	CertFile string
+	KeyFile  string
+}
+
+type fluentbitTLSConfig struct {
+	Enabled    bool
+	SharedKey  string
+	CACertFile string
+	Cert       fluentbitCertConfig
```
pepov: not necessary to split the cert into a separate struct

docs/examples/tls.md
```diff
@@ -0,0 +1,101 @@
...
+apiVersion: logging.banzaicloud.com/v1alpha1
+kind: Fluentd/Fluentbit
+metadata:
+  name: my-fluent-thing
+spec:
+  ...
+  tls:
+    enabled: true
+    secretName: something-something-tls
+    secretType: generic
```
pepov: setting generic should not be required, commented on the type as well

docs/examples/tls.md
```diff
@@ -0,0 +1,101 @@
+
+# TLS Configuration
+
+To configure TLS for fluentd and fluent-bit the operator needs TLS certificates
+to provided via the Custom Resources. This can be done in two ways:
```
pepov:
```suggestion

```

docs/examples/tls.md
```diff
@@ -0,0 +1,101 @@
+
+# TLS Configuration
+
+To configure TLS for fluentd and fluent-bit the operator needs TLS certificates
```
pepov: To configure TLS for `fluentd` and `fluent-bit` the operator needs TLS certificates set via the `Fluentd` and `Fluentbit` Custom Resources respectively. This can be done in two ways:

docs/examples/tls.md
```diff
@@ -0,0 +1,113 @@
+
+# TLS Configuration
+
+To configure TLS for fluentd and fluent-bit the operator needs TLS certificates to provided via the Custom Resources. This can be done in two ways:
```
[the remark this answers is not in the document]
kouk: @pepov did you want to propose instead "To configure ... *you* have to set TLS certificates..."? Otherwise it appears ungrammatical to me.

pkg/apis/logging/v1alpha1/fluentd_types.go
kouk: not entirely sure why we need this but I'll add it anyway. Doesn't this apply only when marshalling to JSON?

docs/examples/tls.md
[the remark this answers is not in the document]
pepov: No I meant it differently, but I can see now what you meant originally. Please simply add 
"to be provided" instead of "to provided"

pkg/apis/logging/v1alpha1/fluentd_types.go
pepov: newer kubernetes versions add schema validation which will fail if this is not set in the yaml (which will translate in the background to json of course)

pkg/resources/fluentd/deployment.go
kouk: @pepov just for uniformity, related to this line:
https://github.com/banzaicloud/logging-operator/pull/134/files#diff-e0cf039a346a2f831decd3a8364eb052R96
A bit OCD of me I would agree. I'll revert it.

pepov: I was too picky perhaps, but thanks!

pepov:
I've just tested this with cert-manager and it worked great, will add the manifests to the examples in master, thanks @kouk !

kouk:
Thanks @pepov for the review and the merge

PR #145, opened 2019-09-20
Mark `path` argument as required

worr:
Mark `path` as required, since it's required for the configuration check to pass

| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | no|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes
| Related tickets | 

### What's in this PR?

Documentation change marking `path` in the `buffer` section required

### Why?

It's required, and the config check will fail to execute without it

### Checklist

- [ ] Implementation tested (with at least one cloud provider)

### DIFF
--- a/docs/plugins/outputs/buffer.md
+++ b/docs/plugins/outputs/buffer.md
@@ -3,7 +3,7 @@
 |---|---|---|---|---|
 | type | string | No | - | Fluentd core bundles memory and file plugins. 3rd party plugins are also available when installed.<br> |
 | tags | string | No |  tag,time | When tag is specified as buffer chunk key, output plugin writes events into chunks separately per tags. <br> |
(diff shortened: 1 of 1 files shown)


pepov:
Docs are fully generated and the reason it's not required is because the `json` tag on the field is marked as `omitempty`.

But I beleive we have a plan to set this automatically to a sane value and just let the user configure it if they need it to be custom. cc @tarokkk 

Does it make sense?

worr:
Ah, my bad, sorry! I didn't realize these docs were auto-generated.

What you're proposing sounds like a great solution.

tarokkk:
The first version will only contain a random value. Later I would like to introduce some plugin specific default: https://github.com/banzaicloud/logging-operator/pull/146/commits/f833b627a15bbf96bc4eabd487af792a2c9a7e37

Issue #154, opened 2019-09-24
Attach plugin secrets to Fluentd Pod


tarokkk:
**Is your feature request related to a problem? Please describe.**
The problem is that many output plugin could require certificates or other file based secrets. As we don't want to recreate Pods on new configuration we need a way to present those secrets runtime.

**Describe the solution you'd like to see**
The idea is to create a generic secret that we are updating runtime so the Pod don't need restarts and the plugin can access to these secrets. The values of secrets can be copied from the original secrets.

PR #166, opened 2019-10-01
Add mountable secrets to Fluentd

tarokkk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #154 

### Checklist


Makefile
```diff
@@ -80,7 +80,7 @@ bin/kubebuilder: bin/kubebuilder_${KUBEBUILDER_VERSION}
 test: generate fmt vet manifests bin/kubebuilder
 	@which kubebuilder
 	@which etcd
-	kubebuilder version
+	bin/kubebuilder version
```
pepov: this was intentional. given the PATH env variable is set to the local bin directory this should work without specifing `bin/` here
```suggestion
	kubebuilder version
```

controllers/logging_controller.go
```diff
@@ -146,6 +148,23 @@ func SetupLoggingWithManager(mgr ctrl.Manager, logger logr.Logger) *ctrl.Builder
...
 				return nil
 			}
+			if o, ok := object.(*corev1.Secret); ok {
+				requestList := []reconcile.Request{}
+				for key, _ := range o.Annotations {
+					r := regexp.MustCompile("logging.banzaicloud.io/(.*)")
+					result := r.FindStringSubmatch(key)
+					if len(result) > 1 {
+						loggingRef := result[1]
+						// When loggingRef is default we trigger "empty" and default loggingRef as well
```
pepov:
```suggestion
						// When loggingRef is default we trigger "empty" and default loggingRef as well, because we can't use an empty string in the annotation, thus we refer to `default` in case the loggingRef is empty
```

docs/plugins/outputs/secret.md
```diff
@@ -19,4 +19,51 @@ For debug purposes you can define secret values directly. However this is *NOT*
...
+```
+<match **>
+    @type forward
+    tls_cert_path /fluentd/etc/secret/default-fluentd-tls-tls.crt
+    ...
+</match>     
+```
+
+### How it works?
+Behind the scene the operator will mark the secret with an annotation. It will watches it for changes as long as the annotation is presented.
```
pepov:
```suggestion
Behind the scene the operator marks the secret with an annotation and watches it for changes as long as the annotation is present.
```

pkg/model/secret/secret.go
```diff
@@ -47,19 +49,81 @@ type KubernetesSecret struct {
...
+	annotationKey := fmt.Sprintf("logging.banzaicloud.io/%s", loggingRef)
+	if k8sSecret.ObjectMeta.Annotations == nil {
+		k8sSecret.ObjectMeta.Annotations = make(map[string]string)
+	}
+	k8sSecret.ObjectMeta.Annotations[annotationKey] = "watched"
+	err = k.client.Update(context.TODO(), k8sSecret)
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("unable to update source secret: %q", secret.MountFrom.SecretKeyRef.Name))
 	}
+	return k.mountConfig.ConfigPath + "/" + secretKey, nil
```
pepov: I think there are some issues with this: we update all the source/target secrets all the time, which will cause a lot of event updates and also it could cause several fluentd reloads as well. The loader should not have any side effects, all the secret preparation should happen in the main reconcile loop. We should move the logic there if possible and also be careful to avoid updating resources if they haven't changed.

Makefile
tarokkk: It doesn't work for me

pkg/model/secret/secret.go
tarokkk: This ha been refactored. The new approach only collects secrets during config generation and reconcile it before the other reconcile cycles.

pkg/resources/fluentd/outputsecret.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+
+	"github.com/banzaicloud/logging-operator/pkg/model/secret"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r *Reconciler) outputSecret(secrets *secret.MountSecrets, mountPath string) runtime.Object {
```
pepov: it would be more clear to return the source secret items from this function and reconcile them outside instead of doing it inline

pkg/resources/fluentd/fluentd.go
```diff
@@ -48,19 +52,26 @@ var labelSelector = map[string]string{
...
 	}
 }
 
 // Reconcile reconciles the fluentd resource
 func (r *Reconciler) Reconcile() (*reconcile.Result, error) {
+	// Prepare output secret
+	err := r.ReconcileResource(r.outputSecret(r.secrets, OutputSecretPath))
+	if err != nil {
+		return nil, errors.WrapIf(err, "failed to reconcile resource")
+	}
```
pepov: is this necessary to do before the configcheck, or could we do it after that where all the other resources are reconciled?

PR #218, opened 2019-11-07
Revamp volume configuration, fix issues introduced in 2.4.0

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
 - Remove the default `hostPath` volumes for fluentbit, introduced by mistake in 2.4.0 and use `emptyDir` instead.
 - Make the path of the `hostPath` volumes automatically configured if not set
 - Remove the default volume-mount-hack and use `fsGroup` instead
 - Add missing `runAsGroup` for container `securityContext`

### Why?
Because hostPath volumes can cause issues on readonly host filesystems.

### Checklist


pkg/resources/fluentd/statefulset.go
```diff
@@ -48,6 +48,23 @@ func (r *Reconciler) statefulset() (runtime.Object, k8sutil.DesiredState) {
 }
 
 func (r *Reconciler) statefulsetSpec() *appsv1.StatefulSetSpec {
+	initContainers := make([]corev1.Container, 0)
+
+	if !r.Logging.Spec.FluentdSpec.DisablePvc && r.Logging.Spec.FluentdSpec.VolumeMountChmod {
```
tarokkk: Do we need the PVC part for this condition?

pkg/sdk/api/v1beta1/common_types.go
```diff
@@ -37,6 +39,41 @@ type Metrics struct {
...
+//           logging system can coexist in case a hostPath mount is used
+// `name`    will be the name of the volume and the lowest level directory in case a hostPath mount is used
+func (storage *KubernetesStorage) GetVolume(logging, name string) corev1.Volume {
+	const (
+		hostPath = "/opt/logging-operator/%s/%s"
+	)
+	volume := corev1.Volume{
+		Name: name,
+	}
+	if storage != nil {
```
tarokkk: We should do this check before invoking the function.

pkg/resources/fluentd/statefulset.go
pepov: not really, I've removed it

pkg/sdk/api/v1beta1/common_types.go
pepov: modified to be a value receiver

PR #250, opened 2019-11-20
fix label handling

### LINKED ISSUE #248: Logging CR labels should not be inherited by lower level components
**Describe the bug**:
Logging CR labels are populated by the helm chart, thus a new release means label changes on the logging component. Currently a change in the logging CR's labels cascades down to all downstream components, that causes problems on some resources that cannot be updated, like statefulsets and daemonsets. Service resource is also problematic, but because of a different issue.

**Expected behaviour**:
Labels should not change on downstream components when the logging labels change. 

Change the labels on the logging resource manually.

**Additional context**:
We will need another issue to deal with any reasonable change required on resources that disallow changing certain fields on the file.

/kind bug

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #248 

### What's in this PR?
Avoid using the label of the Logging CR directly, but make the fluentd/fluentbit related resource labels configurable by the user

QualifiedNamespacedName has been removed due to it was introduced back when the logging resource was namespaced, but not required anymore.

### Why?
Right now every new release - if deployed using helm - causes a change in all the resources' labels which is intended.

### Checklist


worr:
This great, thanks so much!

PR #252, opened 2019-11-21
Manage desired state for resources having limited update capability (statefulset, daemonset)

### LINKED ISSUE #249: Support change management of resources that require object recreation
Whenever we change a field on a resource that cannot be changed, the operator fails and become disfunctional.

Examples:
 - statefulset fields in the spec.template other than replicas and container images.
 - daemonset fields similarly
 
Another issue is that we don't take the existing state of the resources into consideration when creating the desired state and face conflicts when for example updating a `service` object where the `ClusterIP` field is generated by the API Server. In this case the operator sends an update with an empty `ClusterIP` field but that is invalid as it cannot be set by the client. We could argue that the API server should ignore it, but we can't do much with this. Instead we can again take the existing state of the resource into consideration and set the ClusterIP explicitly before updating the object.

To support this we need a more dynamic way of creating the desired objects, where the existing state is also available and where we can mutate the desired object based on that information and where we can decide that an object has to be recreated which is not supported currently

I also recommend to create an opt-in flag for this, so that users may not get surprised by the operator removing the entire daemonset and statefulset because of a change in case of an upgrade. In case a resource has to be recreated the operator should log this with helpful information to the administrator so that they know what kind of flag needs to be set, and that they should configure fluentbit and fluentd with persistent storage (hostPath volume for fluentbit or PVC in case of fluentd  positions and buffers) if they don't want to lose and/or receive duplicated logs.

The same info should be set in the logging CR's status.

No better alternatives currently.

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #249 

### What's in this PR?
Support an for an optional desired state hook to be called before update. It will receive the current state and can alter the desired object if necessary. E.g. service.spec.clusterIP can be set to the current  value to avoid errors when updating with an empty value.

Add support for removing objects if they cannot be updated (similar to how the prometheus operator manages statefulset updates). 

Add and fix some docs.
Fix pvc default config for fluentd.

### Checklist

### DIFF
--- a/pkg/k8sutil/resource.go
+++ b/pkg/k8sutil/resource.go
@@ -16,38 +16,59 @@ package k8sutil
 
 import (
 	"context"
+	"fmt"
 	"reflect"
+	"time"
 
 	"emperror.dev/errors"
 	"github.com/banzaicloud/k8s-objectmatcher/patch"
+	"github.com/banzaicloud/logging-operator/pkg/sdk/api/v1beta1"
 	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/go-logr/logr"
 	"github.com/goph/emperror"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
-	StateAbsent  DesiredState = "Absent"
-	StatePresent DesiredState = "Present"
+	StateAbsent  StaticDesiredState = "Absent"
+	StatePresent StaticDesiredState = "Present"
 )
 
-type DesiredState string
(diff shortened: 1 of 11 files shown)


pkg/k8sutil/resource.go
```diff
@@ -105,25 +132,53 @@ func (r *GenericResourceReconciler) ReconcileResource(desired runtime.Object, de
...
+						)
+						if err != nil {
+							return nil, errors.Wrapf(err, "failed to delete resource %+v", current)
+						}
+						return &reconcile.Result{
+							Requeue:      true,
+							RequeueAfter: time.Second * 10,
+						}, nil
+					} else {
+						return nil, errors.New(
```
worr: Forgive my ignorance here, but will returning an error here cause reconciliation to stop? If so, is it safe to continue instead? If you run into this issue, you block all updates further on in the reconciliation loop, rather than updating what you can and reporting an error for what you cannot update.

pepov: The other option would be to collect all the errors and return with all of them at once at the end of the loop. We could even write the error state into the status field for even better feedback on what is happening. What do you think?

pepov: @worr @ahma Is it ok for you if I stick with the current method for now and create a follow up issue? I would like to do this in a broader context, not just for the reconciler, but for when we have issues with the model as well.

worr: Yeah, I think the current method definitely improves the current situation, regardless of changing how partial updates work. I'd be happy with a follow-up issue.

PR #258, opened 2019-11-25
V1: Permit custom deployment strategy for fluentd

kouk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Added a field `deploymentStrategy` to the fluentd CRD, and add support to the helm chart. The field controls the value of `Spec.Strategy` in the fluentd deployment generated by the operator, which defaults to `RollingUpdate`. The current default behaviour is preserved if the new chart value is left unset, i.e. the CRD spec remains the same. If the value is set, e.g. to `Restart`, then the CRD will be updated accordingly and the fluentd deployment's spec will get `Spec.Strategy` equal to `Restart`.

### Why?
In some providers and storage classes attempting to rollout changes to the fluentd deployment will result in the rollout hanging due to a multi-attach volume error. The solution to this is to terminate the old fluentd pod before starting the new one.

### Checklist


pepov:
@kouk do you prefer a release, or you okay with running from the v1 branch?

kouk:
I'm fine with the branch, since I have other changes in my fork currently.
Hoping to get around to pushing them your way too :-)

PR #269, opened 2019-11-28
multilog to handle fluent logs in a rotated manner instead of writing them to stdout

pepov:
.


tarokkk:
This is outdated due to #270

PR #281, opened 2019-12-06
- Helm3 suppport

### LINKED ISSUE #276: Create a Helm 3 compatible chart for the logging-operator
Helm 3 has removed the crd-install hooks in favour of a crds directory: https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you

The current CRDs for the logging-operator helm chart are located in the templates folder. When using helm 3 to install a helm chart with the logging-operator as a dependency the CRDs are not 
installed before they are needed.

Can we create a Helm 3 compatible chart for the logging operator.

ahma:
- Helm3 support (separated crd install)

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #276 


charts/logging-operator/README.md
```diff
@@ -26,6 +26,10 @@ To install the chart with the release name `my-release`:
 $ helm install --name my-release banzaicloud-stable/logging-operator
 ```
 
+### CRDs
+The CRDs are provisioned using crd-install hooks, rather than relying on a separate chart installation.
```
pepov: In our case this is not true.

charts/logging-operator/README.md
```diff
@@ -26,6 +26,10 @@ To install the chart with the release name `my-release`:
...
+If you already have these CRDs provisioned and don't want to remove them, you can disable the CRD creation by these hooks by passing `createCustomResource=false` (not required if using Helm v3).
```
pepov:
```suggestion
Use `createCustomResource=false` with Helm v3 to avoid trying to create CRDs from the `crds` folder and from templates at the same time.
```

charts/logging-operator/README.md
ahma: Remove thx

PR #289, opened 2019-12-15
raise memory limit to avoid OOM errors when chunks are getting full

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Raise the default memory limit

### Why?
Because when chunks gets full (256M) then the default 200M memory limit will be exhausted.


pehlert:
Hi @pepov, thanks for this fix! This hasn't made it into a release yet unfortunately and we re seeing a lot off oom kills on fluentd.. We're using loki with a file buffer and already tried to adjust chunk sizes etc, but the memory will still grow beyond the allowed 200 megs. Do you have any workaround for this while we wait for the release?

pehlert:
Took me way too long to find.. Sorry!

https://github.com/banzaicloud/logging-operator/blob/master/docs/requirements.md

pepov:
Glad you found it! Hopefully we can cut a new release soon!

PR #298, opened 2019-12-20
FluentD FluentBit liveness, readiness probe config support

### LINKED ISSUE #290: Enable configuring liveness and readiness probes
We are having problems with individual fluentd pods finding themselves detaching all forwarding nodes due to hard timeouts and thus failing to process any more messages. We manually fix that by restarting containers (and are still investigating underlying problems), but would love to be able to specify a liveness check like the one in https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-gcp/fluentd-gcp-ds.yaml#L58 to detect such situations and restart automatically.

Allow configuring a liveness and readiness probe for fluentd (and possibly fluentbit though we currently don't see a use for it).

We are currently doing manually what we hope the liveness probe might trigger for us.

ahma:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #290 

### Checklist


fluentd-image/v1.7/healthy.sh
```diff
@@ -0,0 +1,30 @@
...
+STUCK_THRESHOLD_SECONDS=${STUCK_THRESHOLD_SECONDS:-900};
+
+if [ ! -e ${BUFFER_PATH} ];
+then
+  exit 1;
+fi;
+touch -d "${STUCK_THRESHOLD_SECONDS} seconds ago" /tmp/marker-stuck;
+if [ -z "$(find ${BUFFER_PATH} -type d -newer /tmp/marker-stuck -print -quit)" ];
+then
+  rm -rf ${BUFFER_PATH};
```
pepov: I'm not 100% sure we can do it like this. Looks we can easily lose logs in a scenario where an output is down for 15 mins, which is not normal, but can happen. We should at least make this configurable somehow.

PR #313, opened 2020-01-13
added log_group_name_key attribute

dududko:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Added `log_group_name_key` to CloudwatchOutput struct.

### Why?
This attribute is mentioned in documentation, but it does not present in the code.

### Checklist

- [x ] User guide and development docs updated (if needed)


ahma:
Thanks for your fix @dududko . Could you please run the `make check-diff` command. It would be great too if you could add yourself to the list of adopters.

pepov:
@dududko you should run `make generate manifests` in case there is a change in crds and commit the generated changes as well

pepov:
@dududko could you please also rebase on the latest master?

dududko:
Guys, you need to decide which strategy you are using in this repo, rebase or merge :-)

PR #355, opened 2020-02-10
Add SSL options to Kafka output

velothump:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Adds SSL options to Kafka output

### Why?
My Kafka cluster requires client certs to ship logs.

### Checklist

### DIFF
--- a/pkg/sdk/model/output/kafka.go
+++ b/pkg/sdk/model/output/kafka.go
@@ -93,6 +93,16 @@ type KafkaOutputConfig struct {
 	AckTimeout int `json:"ack_timeout,omitempty"`
 	// The codec the producer uses to compress messages (default: nil). The available options are gzip and snappy.
 	CompressionCodec string `json:"compression_codec,omitempty"`
+	// CA certificate
+	SSLCACert *secret.Secret `json:"ssl_ca_cert,omitempty"`
+	// Client certificate
+	SSLClientCert *secret.Secret `json:"ssl_client_cert,omitempty"`
+	// Client certificate chain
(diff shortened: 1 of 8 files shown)


pkg/sdk/model/output/kafka.go
```diff
@@ -93,6 +93,16 @@ type KafkaOutputConfig struct {
...
+	// CA certificate
+	SSLCACert *secret.Secret `json:"ssl_ca_cert,omitempty"`
+	// Client certificate
+	SSLClientCert *secret.Secret `json:"ssl_client_cert,omitempty"`
+	// Client certificate chain
+	SSLClientCertChain *secret.Secret `json:"ssl_client_cert_chain,omitempty"`
+	// Client certificate key
+	SSLClientCertKey *secret.Secret `json:"ssl_client_cert_key,omitempty"`
+	// Verify certificate hostname
+	SSLVerifyHostname bool `json:"ssl_verify_hostname,omitempty"`
```
pepov: I beleive you won't be able to use this flag to actually disable the ssl hostname verification, since the setting behind this is default true. 

Please check the Elasticsearch plugin to see how this scenario could be implemented. (The only downside is that this configuration flag will be added to the config explicitly even if it's value matches the plugin's default)

velothump: Thanks, I've updated the type to `*bool` which seems to resolve the issue, I've added `ssl_verify_hostname: false` into the test as well.

PR #366, opened 2020-02-17
Add component label to fluentd pods and service selector

eddycharly:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This pull request adds a `component` label to fluentd pods and service selector.

Without the label, the service considers the pod used to check config as a valid target (as the labels of the pod match the service selector).

With the `component` label, only the pods from the fluentd statefullset match the fluentd service.

### DIFF
--- a/pkg/resources/fluentd/appconfigmap.go
+++ b/pkg/resources/fluentd/appconfigmap.go
@@ -39,7 +39,7 @@ func (r *Reconciler) appconfigMap() (runtime.Object, reconciler.DesiredState, er
 	data := make(map[string][]byte)
 	data[AppConfigKey] = []byte(*r.config)
 	return &corev1.Secret{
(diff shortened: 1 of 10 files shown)


pepov:
@eddycharly thanks for pointing this out and for the fix! I think we want to use an officially recommended k8s label for this, just give us a little time to think this through!

eddycharly:
No problem, i was thinking `component` was not perfect too. Maybe something like `app.kubernetes.io/component` ? ;-)

pepov:
exactly and I've just double checked we don't use it anywhere so let's be it!

eddycharly:
Done, i made the component part of the contract, looks cleaner and less error prone.

PR #407, opened 2020-03-18
Loki output extract_kubernetes_labels default false fix

### LINKED ISSUE #403: Output - Loki - extract_kubernetes_labels key is ignored
**Describe the bug**:
Setting `extract_kubernetes_labels` key to false is ignored by the operator. Not even a configcheck is generated.

**Expected behaviour**:
The option to be set to false in the fluentd config.

On a Loki output set `extract_kubernetes_labels` to false. Get the fluentd config or the configcheck pod's logs and the key will be true. Logging operator doesn't notice the change if you re-apply the Output with different values for that property.

**Additional context**:
The property is described [here](https://github.com/banzaicloud/logging-operator/blob/3.0.0/docs/plugins/outputs/loki.md#configuration).

**Environment details**:
- Kubernetes version (e.g. v1.15.2): 1.16
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): on-premise
- logging-operator version (e.g. 2.1.1): 3.0.0

/kind bug

ahma:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #403

### Checklist

### DIFF
--- a/pkg/sdk/model/output/loki.go
+++ b/pkg/sdk/model/output/loki.go
@@ -64,7 +64,7 @@ type LokiOutput struct {
 	// Format to use when flattening the record to a log line: json, key_value (default: key_value)
 	LineFormat string `json:"line_format,omitempty" plugin:"default:json"`
 	// Extract kubernetes labels as loki labels (default: false)
-	ExtractKubernetesLabels *bool `json:"extract_kubernetes_labels,omitempty" plugin:"default:false"`
+	ExtractKubernetesLabels *bool `json:"extract_kubernetes_labels,omitempty"`
 	// Comma separated list of needless record keys to remove (default: [])
 	RemoveKeys []string `json:"remove_keys,omitempty"`
 	// If a record only has 1 key, then just set the log line to the value and discard the key. (default: false)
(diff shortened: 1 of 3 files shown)


pkg/sdk/model/output/loki.go
```diff
@@ -64,7 +64,7 @@ type LokiOutput struct {
 	// Format to use when flattening the record to a log line: json, key_value (default: key_value)
 	LineFormat string `json:"line_format,omitempty" plugin:"default:json"`
 	// Extract kubernetes labels as loki labels (default: false)
-	ExtractKubernetesLabels bool `json:"extract_kubernetes_labels,omitempty"`
+	ExtractKubernetesLabels *bool `json:"extract_kubernetes_labels,omitempty" plugin:"default:false"`
```
pepov:
```suggestion
	ExtractKubernetesLabels *bool `json:"extract_kubernetes_labels,omitempty"`
```

pkg/sdk/model/output/loki.go
```diff
@@ -125,7 +125,7 @@ func (l *LokiOutput) ToDirective(secretLoader secret.SecretLoader, id string) (t
 		} else {
 			l.RemoveKeys = []string{"kubernetes"}
 		}
-		l.ExtractKubernetesLabels = true
+		l.ExtractKubernetesLabels = util.BoolPointer(true)
```
pepov:
```suggestion
		if l.ExtractKubernetesLabels == nil {
			l.ExtractKubernetesLabels = util.BoolPointer(true)
		}
```

ahma: Thanks

PR #419, opened 2020-03-26
Fix and unify IDs

tarokkk:
Unify IDs and metric names

### DIFF
--- a/pkg/resources/model/system.go
+++ b/pkg/resources/model/system.go
@@ -16,6 +16,7 @@ package model
 
 import (
 	"fmt"
+	"strconv"
 
 	"emperror.dev/errors"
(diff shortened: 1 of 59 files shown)


pkg/resources/model/system.go
```diff
@@ -223,7 +229,8 @@ func FlowDispatcher(flowCr interface{}) (*CommonFlow, error) {
 				},
 			}
 		}
-		flow, err := types.NewFlow(matches, f.Name, f.Namespace)
+		id := fmt.Sprintf("flow" + ":" + f.Namespace + ":" + f.Name)
```
siliconbrain: Either use `fmt.Sprintf("flow:%s:%s", f.Namespace, f.Name)` or just `"flow" + ":" + f.Namespace + ":" + f.Name`

PR #424, opened 2020-03-30
Configure default_route option from fluent-plugin-label-router

### LINKED ISSUE #423: Configure default_route option from fluent-plugin-label-router
I'd like to define a Flow to catch all logs that have not matched any other Flow or ClusterFlow. This is so every stdout line from containers can be handled without duplication.

Specifically, I'm working on a multi-tenant Kubernetes cluster and want to provide an Elasticsearch index of unhandled logs. If a team wishes for their logs to be handled differently (for example the logs contain sensitive data and need more restrictive access control) the team can define a Flow and Output object in their own namespace.

I could define a ClusterFlow without any `match` arguments. However that ClusterFlow also matches logs that have already been matched by other Flow and ClusterFlows.

There are of options I can imagine. Either:
Expose a DefaultClusterFlow field on the LoggingSpec.
OR
Expose a DefaultClusterFlow type. Only one instance of DefaultClusterFlow is allowed per Logging instance. The DefaultClusterFlowSpec looks like a ClusterFlowSpec but it doesn't contain a Match or Selectors field.

**Pros of the new field**
Doesn't introduce another Flow variant object.
More intuitive at-most-one definition of a default flow. A user can either set the field or not, they cannot accidentally define more than one default flows.

**Cons of the new field**
The referenced ClusterFlow has two purposes now. One to process logs according to some Match or Selectors and one to process logs that didn't match any other ClusterFlow/Flow. This can be non-intuitive. Furthermore, the Logging object has a weak reference to the ClusterFlow name. If the ClusterFlow object is deleted or renamed, the reference would still exist.

**Pros of the new type**
This Flow type has a clear purpose since it doesn't have a Match or Selectors field. 

**Cons of the new type**
Yet another Flow variant, I still wonder if this confusing to users.

PhilGrayson:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #423

### What's in this PR?
This is a basic implementation of DefaultClusterFlow type option described in issue #423. 

This code is only a draft and this PR is mainly to gauge interest in the feature request from #423. Tests and documentation are purposefully missing until there is consensus that Banzai Cloud wishes to support this feature.

Applying the following manifests:
```
cat <<EOF | kubectl -n logging apply -f -
kind: Logging
metadata:
  name: shared
spec:
  fluentd: {}
  fluentbit: {}
  controlNamespace: logging
  defaultFlow:
    outputRefs:
      - some-output
EOF

cat <<EOF | kubectl -n logging apply -f -
kind: ClusterOutput
metadata:
  name: some-output
spec:
  nullout: {}
EOF
```

Renders the following fluentd config:
```
<match **>
  @type label_router
  @id main_label_router
  default_route @9e81e7b963c71363e2fb3eefcfecfc0e
</match>
<label @9e81e7b963c71363e2fb3eefcfecfc0e>
  <match **>
    @type null
    @id _shared_null-output_null
  </match>
</label>
```

### Checklist

### To Do
- [ ] Add tests
- [ ] Update documentation


controllers/logging_controller.go
```diff
@@ -263,6 +265,28 @@ func (r *LoggingReconciler) GetResources(logging *loggingv1beta1.Logging) (*mode
...
+		return nil, err
+	}
+
+	if len(defaultClusterFlows.Items) > 0 {
+		for _, i := range defaultClusterFlows.Items {
+			if i.Spec.LoggingRef == logging.Spec.LoggingRef {
+				if defaultClusterFlow != nil {
+					return nil, errors.New("there can only be one DefaultClusterFlow per Logging")
+				}
+				defaultClusterFlow = &i
```
PhilGrayson: This kind of validation would suit a validating admission webhooks

pkg/resources/model/system.go
```diff
@@ -164,10 +175,24 @@ func GetFlowMatchFromSpec(namespace string, matches interface{}) ([]types.FlowMa
...
 }
 
-func FlowDispatcher(flowCr interface{}) (*CommonFlow, error) {
+func (l *LoggingResources) FlowDispatcher(flowCr interface{}) (*CommonFlow, error) {
 	var commonFlow *CommonFlow
 	var err error
 	switch f := flowCr.(type) {
+	case *v1beta1.DefaultFlowSpec:
+		id := fmt.Sprintf("logging:%s:%s", l.logging.Namespace, l.logging.Name)
+		flow, err := types.NewFlow([]types.FlowMatch{}, id, l.logging.Name, l.logging.Namespace)
```
pepov, line 184: I'm not sure this would work without an actual flow match defined. I would need to create a test case for this, but without digging deeper, I think we just need an empty FlowMatch item put in the list.

pepov: nevermind, I didn't think this through carefully

tarokkk:
I just checked the code and tested it locally. Nice work ;) Because the `label_router`'s feature`default_route` hasn't been tested in a production environment there could be bugs but the configuration generation seems perfect.

PhilGrayson:
Brilliant, thank you for testing and merging.

PR #460, opened 2020-04-25
http output plugin

osela:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Added support for fluentd's http output plugin (https://docs.fluentd.org/output/http).

### Additional context

Added support for `[]int` fields in `StructToStringMapper` for the `retryable_response_codes` field.

### Checklist


pkg/sdk/model/output/http.go
```diff
@@ -0,0 +1,147 @@
...
+	// The client certificate path for TLS.
+	TlsClientCertPath *secret.Secret `json:"tls_client_cert_path,omitempty"`
+	// The client private key path for TLS.
+	TlsPrivateKeyPath *secret.Secret `json:"tls_private_key_path,omitempty"`
+	// The client private key passphrase for TLS.
+	TlsPrivateKeyPassphrase *secret.Secret `json:"tls_private_key_passphrase,omitempty"`
+	// The verify mode of TLS. [peer, none] (default: peer)
+	TlsVerifyMode string `json:"tls_verify_mode,omitempty"`
+	// Raise UnrecoverableError when the response code is non success, 1xx/3xx/4xx/5xx. If false, the plugin logs error message instead of raising UnrecoverableError. (default: true)
+	ErrorResponseAsUnrecoverable bool `json:"error_response_as_unrecoverable,omitempty"`
```
pepov, line 85:
```suggestion
	ErrorResponseAsUnrecoverable *bool `json:"error_response_as_unrecoverable,omitempty"`
```
so that a false value won't be cleared by omitempty

pkg/sdk/model/output/http.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// +kubebuilder:object:generate=true
+// +docName:"HTTP auth config"
+// http_auth
+type HTTPAuth struct {
+	// Username for basic authentication.
+	Username string `json:"username"`
+	// Password for basic authentication.
+	Password string `json:"password"`
```
pepov, line 140: This should be a `secret.Secret` I beleive.

osela: done. notice that these fields are actually required, but the current implementation doesn't recursively handle tags like `plugin:"required"` for struct field and I didn't want to dive too deep into a major refactor.

PR #494, opened 2020-06-02
Fix helm chart to properly set clusterIP

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Fix http.service.clusterIP value to mind the scoping and to be backwards compatible with 3.0.5 and earlier chart versions

### Why?
Charts prior to 3.0.5 (inclusive) used `ClusterIP: None` which was then changed to `ClusterIP: ""` which is bad for two reasons:
- it is backwards incompatible
- it behaves really bad with helm3, which tries to update this value on each upgrade to set it back to the empty string which fails on the API server

### DIFF
--- a/charts/logging-operator/Chart.yaml
+++ b/charts/logging-operator/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "3.2.0"
 description: A Helm chart to install Banzai Cloud logging-operator
 name: logging-operator
-version: 3.2.1
+version: 3.2.2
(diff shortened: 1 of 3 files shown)


pepov:
@velothump you may want to defend why the `clusterIP` change from `None` to the empty string was required or just a typo in this commit: https://github.com/banzaicloud/logging-operator/commit/c1b192b

velothump:
Sorry about that, think it was just a copy and paste error. :(

Issue #497, opened 2020-06-14
Let ClusterFlow and ClusterOutput defined in arbitrary namespaces


pepov:
**Is your feature request related to a problem? Please describe.**
A team that has to deploy cluster wide logging rules - without having access to the logging's control namespace - currently has no good options since ClusterFlow and ClusterOutput are namespaced resources.

**Describe solution and alternatives you've considered**
They would either define ClusterFlow and ClusterOutput as cluster scoped resources, but that would require these resource types to be changed which would be backwards incompatible with older clients.

The other option would be to make the ClusterFlow and ClusterOutput resources available for the logging system from any/all namespaces.

PR #499, opened 2020-06-15
Allow ClusterOutput and ClusterFlow resources defined in arbitrary namespaces

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #497

### What's in this PR?
Adds an optional, backwards compatible flag to allow using `ClusterOutput` and `ClusterFlow` resources in arbitrary namespaces. Administrators will still be able to control who has the ability to create such resources through RBAC.

### Why?
To support use cases where a development team has ownership over logs coming from multiple namespaces but they can only deploy resources into their own namespace and cannot access the logging's control namespace.


controllers/logging_controller.go
```diff
@@ -263,8 +263,14 @@ func (r *LoggingReconciler) GetResources(logging *loggingv1beta1.Logging) (*mode
 	loggingResources := model.NewLoggingResources(logging, r.Client, r.Log)
 	var err error
 
+	listOptsForClusterResources := []client.ListOption{}
```
siliconbrain, line 266:
```suggestion
	var listOptsForClusterResources []client.ListOption
```
Could save an unnecessary allocation.

Issue #439, opened 2020-04-08
Separate Output and ClusterOutput references for more clarity


pepov:
Currently `Output` and `ClusterOutput` objects are referenced ambigously from `Flow` and `ClusterFlow` resources. 

In case of a `Flow` resource both an `Output` and `ClusterOutput` can be referenced the same way. If they have the same name an `Output` will take priority and will shadow the `ClusterOutput`.

In case of a `ClusterFlow` only a `ClusterOutput` can be referenced, but for someone not intimate with this rule is not fully obvious what will happen.

To make this situation clear the reference should be unambigous and `outputRefs` should refer only to namespaced outputs, whereas a new `clusteroutputRefs` field in the `Flow`'s spec should refer to `ClusterOutput` objects only.

tarokkk:
Absolutely agreed! There were several confusions from the users as well, maybe this will clarify the purpose behind those resource types.

pepov:
With https://github.com/banzaicloud/logging-operator/pull/499 now merged we should add namespaced references to clusteroutputs to avoid ambiguity

PR #550, opened 2020-08-10
Separate cluster output refs from output refs

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | no
| Related tickets | #439

### What's in this PR?
As discussed, instead of the `ClusterOutputRefs` field mentioned in the issue, the `GlobalOutputRefs` and `LocalOutputRefs` fields are added to CRDs to make downgrading possible without loss of data. The controller's logic is adjusted to use the new fields without checking the deprecated `OutputRefs` field, thus requiring the CRs to be updated manually.

### Why?
See #439


pkg/resources/model/system.go
```diff
@@ -281,46 +283,49 @@ func (l *LoggingResources) CreateFlowFromCustomResource(flowCr interface{}) (*ty
...
 		flowType = "flow"
 	} else {
 		flowType = "clusterflow"
 	}
 	var multierr error
+
 FindOutputForAllRefs:
 	for _, outputRef := range commonFlow.OutputRefs {
-		// only namespaced flows should use namespaced outputs
-		if commonFlow.Scope != "" {
```
pepov, line 293: Is the intent here to allow clusterflows accessing namespaced outputs?

siliconbrain: No, why would you think that?

pepov: Because the condition has been removed, and I can't see what replaces it. (It's not necessarily a bad idea, just not something I would have expected.)

siliconbrain: The check has been replaced by [`FlowDispatcher` assigning `ClusterFlow`s' `OutputRefs` to `CommonFlow`s' `ClusterOutputRefs`](https://github.com/banzaicloud/logging-operator/pull/550/files#diff-a9c7f29e4214f03e1e85112070db41bfR211).

In a way, you're right, cluster flows _could_ also reference namespaced outputs (if their CRDs were extended and the `ClusterFlow`→`CommonFlow` conversion logic adjusted accordingly), but only from their own namespace, which is the control namespace — and while cluster flows are still namespaced resources, which is not for long.

tarokkk (approved):

tarokkk (approved):

ahma (approved):
LGTM

PR #570, opened 2020-09-02
add color stripper plugin

vhdirk:
| Q               | A
| --------------- | ---
| Bug fix?        |no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | no

### What's in this PR?
Adds the [Fluent Color Stripper plugin](https://github.com/mattheworiordan/fluent-plugin-color-stripper)

### Why?
I have several pods that output logs in color. And I can't change that. So Color Stripper to the rescue

### Checklist


tarokkk:
The problem with this plugin that it uses Tag to mark records already stripped. This filter would be much better as a `filter` plugin not as an `output` plugin. There is no reason to define an output plugin for this case. And we have to be careful with such plugins because they reemit logs and possibly have side effects (like tag_normaliser). My suggestion is to rework this plugin as filter (it's really easy just rename the method and filter type) and we can include it in the operator.

vhdirk:
I'm not the author of the filter itself, so I'll pass on the comment. Thanks!

tarokkk:
No, sadly it means to rewrite the fluentd plugin in the correct way.

pepov:
@vhdirk do you have plans to move this forward on the plugin's side? If yes, feel free to reopen, if not sorry that this didn't work out!

jawabuu:
@vhdirk @pepov 
Hoped this would have been completed.
Meanwhile for the benefit of others, I'm currently using this

```
  filters:
    - record_transformer:
        enable_ruby: true
        records:
        - message: ${record["message"].gsub(/\033\[\d{1,2}(;\d{1,2}){0,2}[mGK]/, '')
```

busyboy77:
@jawabuu -- missing '}' in your snippet.

```
  filters:
    - record_transformer:
        enable_ruby: true
        records:
        - message: ${record["message"].gsub(/\033\[\d{1,2}(;\d{1,2}){0,2}[mGK]/, '') }
``` 


secondly, I'm getting like 

```
2024-08-01 17:34:02 +0000 [warn]: #0 send an error event to @ERROR: error_class=RuntimeError error="failed to expand `record[\"message\"].gsub(/\\033\\[\\d{1,2}(;\\d{1,2}){0,2}m/, '')` : error = undefined method `gsub' for nil:NilClass" location="/usr/local/bundle/gems/fluentd-1.16.5/lib/fluent/plugin/filter_record_transformer.rb:310:in `rescue in expand'" tag="kubernetes.var.log.containers.my-container-0e384a513acb6b7376e11481d8449923c1dd9944336c2902cded5bcaae84b078.log" time=2024-08-01 17:34:02.006874082 +0000

```

PR #593, opened 2020-10-14
Remove extra, useless `-d` flag in liveness check

worr:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This removes a spurious `-d` flag in the `touch` command in fluentd's liveness check.

### Why?

When you pass multiple of the same flag to `busybox`, only the last one is used.

```bash
/ # touch /tmp/marker-liveness 
/ # ls -lh /tmp/marker-liveness 
-rw-r--r--    1 root     root           0 Oct 14 14:39 /tmp/marker-liveness
/ # touch -d any -d bullshit -d whatsoever -d @1602683841 /tmp/marker-liveness 
/ # ls -lh /tmp/marker-liveness 
-rw-r--r--    1 root     root           0 Oct 14 13:57 /tmp/marker-liveness
```

Removing it adds clarity and removes confusion.

### DIFF
--- a/fluentd-image/v1.10/healthy.sh
+++ b/fluentd-image/v1.10/healthy.sh
@@ -16,7 +16,7 @@ if [ ! -e ${BUFFER_PATH} ];
 then
   exit 1;
 fi;
-touch -d date -d "@$(($(date +%s) - $LIVENESS_THRESHOLD_SECONDS))" /tmp/marker-liveness;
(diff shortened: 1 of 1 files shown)


tarokkk (approved):

ahma (approved):
Thanks @worr LGTM

PR #606, opened 2020-10-22
Use a more resilient s3 object name by default

worr:
| Q               | A
| --------------- | ---
| Bug fix?        | kinda
| New feature?    | kinda
| API breaks?     | kinda
| Deprecations?   | no

### What's in this PR?
This changes the default `s3_object_key_format`.

### Why?

The s3 plugin uses a default object key that is problematic in a few ways.

1. It makes HEAD requests for each chunk it uploads, starting from 1 each time.
If you have uploaded 2000 log files within the same time slice, it will make
2001 HEAD requests to figure out if it exists.

https://github.com/fluent/fluent-plugin-s3/issues/160

2. The above check is not thread-safe, and two threads can race and decide to
use the same `%{index}` value, with the loser of the race overwriting the chunk
from the winner.

https://github.com/fluent/fluent-plugin-s3/issues/326

This is planned to change for v2, but there's no clear path to v2 right now.
The plugin does warn already if you use multiple threads and don't use either
`%{chunk_id}` or `%{uuid_hash}` in the object key.

### Additional context

This does kind of break the API of the logging-operator, in that the final names of artifacts will be different after this is merged. Given that the default object key format is not super descriptive, I'm assuming that people that may need more explicit names have already overridden this variable.

This does preserve ordering of keys by putting the uuid after the `%{time_slice}`, in the case that people use the time-based keys to determine when a log was uploaded rather than any metadata.

I chose `%{uuid_hash}` over `%{chunk_id}` because I'm a little wary to choose a string that may or may not have strong uniqueness guarantees. Despite a `%{random_hex}` variable, this still triggers a warning in the plugin.

This seems like a good trade-off between the least impact while mitigating a common pitfall that users can easily run into (we ran into this which caused major issues in our logging infrastructure, between incredibly high latency as well as increased s3 costs due to requests and access logging).


tarokkk:
We introduced a similar concept with the One Eye format in PR #609 . I hope this resolves this PR as well.

worr:
@tarokkk Looks like it does. Thanks so much!!

PR #621, opened 2020-11-13
Add support for net config options in fluentbit

worr:
fluentbit offers a few very useful network-related config tunables in the `net`
namespace of their configuration.

One of the more useful for active kubernetes clusters is the new
`net.keepalive_max_recycle` option, which will discard keepalive connections
after they've sent a configured number of messages (added in 1.6.2:
https://fluentbit.io/announcements/v1.6.2/).

Under the old behavior, an active kubernetes cluster could have extremely
long-living fluentbit -> fluentd connections, leading to uneven load on fluentd
instances. The `net.keepalive_max_recycle` option can be used to cycle these
connections so that we keep the distribution of fluentbit's to fluentd's more
even.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
As explained in the commit message, this exposes fluentbit net options, and bumps the version of fluentbit so that we have a very useful network option exposed to us.

### Why?
We had found in our high-volume clusters that load to fluentd's was very lopsided after a long period of time. While this could be mitigated by bouncing all fluentbits, exposing useful network options makes it a lot easier to prevent this through configuration.

### Checklist

### To Do

More documentation *could* be added to the helm charts. However, since I didn't see a place for documentation for parameters like `positiondb` or `bufferStorageVolume`, I've followed suit with the `net` options. If necessary, I can provide more documentation.


pepov:
Hey, nice addition! I think if we document default values, then we can simply set them in case the field is nil and avoid the XYZset boolean helper variables.

worr:
I tried to avoid doing that specifically because theoretically, fluent-bit can change their defaults at any time, so defaults set here could lag behind their changes.

The flipside of this is that we could benefit from setting more kubernetes-specific defaults, like setting `net.keepalive_max_recycle`.

For now, I'll set the current fluent-bit defaults.

pepov:
My point was that if we document defaults explicitly we may drift away actual fluentbit defaults over time. I'm not against either options anyways.

worr:
Ahh, sorry, I misread. I'll leave the diff as-is then.

pepov (approved):

ahma (approved):
Great job thanks!

PR #646, opened 2020-11-30
Fluentd multi process workers config

### LINKED ISSUE #631: Enable root_dir configuration when Worker is > 1 and disable buffer path
https://docs.fluentd.org/deployment/multi-process-workers

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #631

### What's in this PR?
Made `root_dir` configurable via `logging.spec.fluentd.rootDir`. Also, `buffer` directives' `path` parameters are removed if Fluentd is configured to use more than 1 workers. (I'm not proud about how the removal is implemented, but it works, and should be _easily_ improved after **_some_** refactoring.)


pepov (approved):
nit: to make the test less brittle, could we match for the buffer section only?

siliconbrain:
> nit: to make the test less brittle, could we match for the buffer section only?

Added some regexp to make check more specific.

tarokkk (approved):

tarokkk (approved):

PR #665, opened 2021-01-05
Feature/multi fluentbit support

ahma:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### Checklist


pkg/sdk/api/v1beta1/node_agent_types.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/banzaicloud/operator-tools/pkg/typeoverride"
+	"github.com/banzaicloud/operator-tools/pkg/types"
+	"github.com/banzaicloud/operator-tools/pkg/volume"
+)
+
+// +kubebuilder:object:generate=true
+
+type NodeAgent struct {
+	Name          string              `json:"name,omitempty"`
+	Type          string              `json:"type,omitempty"`
```
pepov, line 27: We must emphasize here and in the docs that based on the given type, the operator will automatically configure the required nodeselectors and tolerations for the workload.

We need to highlight the fact that a fully custom configuration is also supported.

pkg/resources/nodeagent/daemonset.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (n *nodeAgentInstance) daemonSet() (runtime.Object, reconciler.DesiredState, error) {
+	var containerPorts []corev1.ContainerPort
+	if n.nodeAgent.FluentbitSpec.Metrics != nil && n.nodeAgent.FluentbitSpec.Metrics.Port != 0 {
+		containerPorts = append(containerPorts, corev1.ContainerPort{
+			Name:          "monitor",
+			ContainerPort: n.nodeAgent.FluentbitSpec.Metrics.Port,
+			Protocol:      corev1.ProtocolTCP,
+		})
+	}
```
nickgerace, line 47: Just a question: these ports are only required for the metrics server?

pkg/resources/nodeagent/nodeagent.go
```diff
@@ -0,0 +1,395 @@
...
+				DB:              util.StringPointer("/tail-db/tail-containers-state.db"),
+				MemBufLimit:     "5MB",
+				Tag:             "kubernetes.*",
+			},
+			Security: &v1beta1.Security{
+				RoleBasedAccessControlCreate: util.BoolPointer(true),
+				SecurityContext:              &v1.SecurityContext{},
+				PodSecurityContext:           &v1.PodSecurityContext{},
+			},
+			ContainersPath: "/var/lib/docker/containers",
```
nickgerace, line 99: Is this going to use the `mountPath` field from the `fluentbitSpec` eventually?

pkg/resources/nodeagent/nodeagent.go
```diff
@@ -0,0 +1,395 @@
...
+		},
+		ContainersPath: "C:\\ProgramData\\docker",
+		VarLogsPath:    "C:\\var\\log",
+		DaemonSetOverrides: &typeoverride.DaemonSet{
+			Spec: typeoverride.DaemonSetSpec{
+				Template: typeoverride.PodTemplateSpec{
+					Spec: typeoverride.PodSpec{
+						Containers: []v1.Container{
+							{
+								Name:    containerName,
```
nickgerace, line 217: I have two questions here.
1. Is this going to be exposed via `values.yaml`?
2. Is the Dockerfile for this going to be included in this PR?
CC: @cbron

nickgerace (commented):
Thanks for doing this again! Had a few comments.

pkg/resources/nodeagent/daemonset.go
ahma: Yes, and also for the liveness check

pkg/resources/nodeagent/nodeagent.go
pepov: 1. yes
2. yes, we will include it as an example, just won't setup any automated build for it. once you have a build and image in place we can use and add it as the default windows image

nickgerace: Okay great. I'll get started on that: https://github.com/rancher/fluent-bit-package

pkg/resources/nodeagent/daemonset.go
nickgerace: Ok great. Thank you

fluent-bit-image/v1.6/Dockerfile.windows
```diff
@@ -0,0 +1,69 @@
...
+
+ARG WINDOWS_VERSION=ltsc2019
+
+#
+# Builder Image - Windows Server Core
+#
+FROM mcr.microsoft.com/windows/servercore:$WINDOWS_VERSION as builder
+
+# The FLUENTBIT_VERSION ARG must be after the FROM instruction
+ARG FLUENTBIT_VERSION=1.7.0
```
nickgerace, line 11: @ahma should this directory be called `v1.7` since the Dockerfile uses Fluentbit 1.7?

nickgerace (approved):

charts/logging-operator-logging/templates/logging.yaml
```diff
@@ -54,4 +54,9 @@ spec:
   {{- else }}
   fluentbit: {}
   {{- end }}
-
+{{- if .Values.nodeAgents }}
+{{ toYaml .Values.nodeAgents | indent 4 }}
+    {{- end }}
+{{- else }}
+    nodeAgents: {}
+{{- end }}
```
nickgerace, line 62: Wouldn't we be missing the `nodeAgents:` field for this? Would it be better to do the following?

```yaml
{{- with .Values.nodeAgents }}
nodeAgents:
  {{- toYaml . | nindent 4 }}
{{- end }}
```

nickgerace (changes requested):

nickgerace:
Seems like there may be a helm chart build error. Hmmm I am unable to see, otherwise I would help :)
https://github.com/banzaicloud/logging-operator/pull/665/checks?check_run_id=2249718349

tarokkk (approved):

charts/logging-operator-logging/templates/logging.yaml
```diff
@@ -54,4 +54,9 @@ spec:
   {{- else }}
   fluentbit: {}
   {{- end }}
-
+{{- if .Values.nodeAgents }}
+{{ toYaml .Values.nodeAgents | indent 4 }}
+    {{- end }}
```
tarokkk, line 59:
```suggestion
```

ahma: Thanks!

PR #666, opened 2021-01-06
Assume Helm 3 as the default

pepov:
Do not install CRDs from templates.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
Switch to assuming helm 3 by default when installing the chart and do not install CRDs as templates, but rely on the crds folder to install them.

### Why?
Because otherwise helm3 will fail when trying to install crds from templates, since those have already been installed in the beginning from the crds folder.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,9 +4,13 @@
 version: 2.1
 
 orbs:
-  helm: banzaicloud/helm@0.0.7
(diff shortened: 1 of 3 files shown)


ahma (approved):
lgtm

siliconbrain (approved):
lgtm

PR #719, opened 2021-04-19
FluentBit  1.7.3

ahma:
- fluentbit 1.6.10 - 1.7.3
- kubelet metadata support

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### Checklist


pepov (approved):

pkg/resources/fluentbit/configsecret.go
```diff
@@ -201,6 +202,10 @@ func (r *Reconciler) configSecret() (runtime.Object, reconciler.DesiredState, er
 		}
 	}
 
+	if r.Logging.Spec.FluentbitSpec.FilterKubernetes.BufferSize != "0" && r.Logging.Spec.FluentbitSpec.FilterKubernetes.BufferSize != "" {
```
pepov, line 205: Empty string means that fluentbit will use the default, which is 32k, which is not enough, so in that case I beleive we need to set it to 0 explicitly when we set the defaults.

pepov (approved):

PR #721, opened 2021-04-21
Make SAS Token and Access Key optional for Azure Blob Storage Output

EppO:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | no

### What's in this PR?

This PR makes both Access Key and SAS token optional values for the Azure Blob Storage Output

### Why?

Reading the [Microsoft fluentd plugin for Azure Blob Storage](https://github.com/microsoft/fluent-plugin-azure-storage-append-blob), `azure_storage_access_key` and `azure_storage_sas_token` both need to be empty when using MSI authentication. If not using MSI, either `azure_storage_access_key` or `azure_storage_sas_token` need to be set, not both (they are just a different methods to authenticate against the Azure Storage Account).

### Additional context

Official documentation from the Azure Blob Storage Output plugin for fluentd:
```<match pattern>
  type azure-storage-append-blob

  azure_storage_account             <your azure storage account>
  azure_storage_access_key          <your azure storage access key> # leave empty to use MSI
  azure_storage_sas_token           <your azure storage sas token> # leave empty to use MSI
  azure_imds_api_version            <Azure Instance Metadata Service API Version> # only used for MSI
  azure_token_refresh_interval      <refresh interval in min> # only used for MSI
  azure_container                   <your azure storage container>
  auto_create_container             true
  path                              logs/
  azure_object_key_format           %{path}%{time_slice}_%{index}.log
  time_slice_format                 %Y%m%d-%H
  # if you want to use %{tag} or %Y/%m/%d/ like syntax in path / azure_blob_name_format,
  # need to specify tag for %{tag} and time for %Y/%m/%d in <buffer> argument.
  <buffer tag,time>
    @type file
    path /var/log/fluent/azurestorageappendblob
    timekey 120 # 2 minutes
    timekey_wait 60
    timekey_use_utc true # use utc
  </buffer>
</match>
```

### Checklist

- [x] User guide and development docs updated (if needed). Looks like `make docs` is not required anymore? 


ahma (approved):
LGTM, Thanks @EppO

PR #756, opened 2021-06-08
Drain review - doc

pepov:
Add design and manual test setup doc. Test steps are still missing.

### DIFF
--- a/docs/volume-drainer.md
+++ b/docs/volume-drainer.md
@@ -5,51 +5,62 @@
 Scaling out fluentd is trivial as long as the sender can distribute load evenly across fluentd instances.
 In Kubernetes this is achieved using a standard service.
 
-However scaling it back is nontrivial because a stopped fluentd pod will not necessarily flush all it's buffers.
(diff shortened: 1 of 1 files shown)


docs/volume-drainer.md
```diff
@@ -0,0 +1,187 @@
...
+kubectl exec log-target -- curl -sS http://localhost:8081/on
+```
+
+Buffers should start disappearing after a while.
+When all buffers are gone, the drainer job should be deleted along with its pod(s) and the placeholder pod, and the PVC `drainer-fluentd-buffer-drainer-fluentd-1` should be marked as drained.
+
+After the placeholder pod is gone, the statefulset should be able to start all replicas.
+(The retry backoff might delay the creation of the pods a bit.)
+
+After the statefulset's pods are all ready, the previously drained PVC should lose its `drained` label.
```
pepov, line 187: ### Clean it up

```
one-eye log-generator delete --update
k delete logging-all --all
k delete po log-target

kind create cluster --name drainer
```

PR #757, opened 2021-06-15
Controller runtime v0.9.0

bshifter:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Prerequisite for custom caching for watchers.

### Checklist


Makefile
```diff
@@ -144,22 +137,14 @@ bin/golangci-lint_${GOLANGCI_VERSION}: | bin
...
-
-bin/kubectl: | bin/kubebuilder_${KUBEBUILDER_VERSION} bin
-	ln -sf kubebuilder_${KUBEBUILDER_VERSION}/kubectl $@
-
-bin/kubebuilder_${KUBEBUILDER_VERSION}: | bin
-	find $(PWD)/bin -name 'kubebuilder_*' -exec rm -r {} +
-	curl -L https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${KUBEBUILDER_VERSION}/kubebuilder_${KUBEBUILDER_VERSION}_${OS}_amd64.tar.gz | tar xvz -C bin
-	ln -sf kubebuilder_${KUBEBUILDER_VERSION}_${OS}_amd64/bin $@
+# Run tests
+ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
```
siliconbrain, line 141:
```suggestion
ENVTEST_ASSETS_DIR=${PWD}/testbin
```

pepov (approved):

Makefile
```diff
@@ -7,15 +8,14 @@ CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false,maxDescLen=
 
 CONTROLLER_GEN_VERSION = v0.5.0
 GOLANGCI_VERSION = v1.33.0
-KUBEBUILDER_VERSION = 2.3.1
 LICENSEI_VERSION = v0.3.1
+ENVTEST_CTRL_VERSION = v0.8.3
```
siliconbrain, line 12: why not use latest version (0.9.0)?

Makefile
```diff
@@ -144,22 +137,14 @@ bin/golangci-lint_${GOLANGCI_VERSION}: | bin
...
-	ln -sf kubebuilder_${KUBEBUILDER_VERSION}/kubectl $@
-
-bin/kubebuilder_${KUBEBUILDER_VERSION}: | bin
-	find $(PWD)/bin -name 'kubebuilder_*' -exec rm -r {} +
-	curl -L https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${KUBEBUILDER_VERSION}/kubebuilder_${KUBEBUILDER_VERSION}_${OS}_amd64.tar.gz | tar xvz -C bin
-	ln -sf kubebuilder_${KUBEBUILDER_VERSION}_${OS}_amd64/bin $@
+# Run tests
+ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
+test: generate fmt vet manifests
+	mkdir -p ${ENVTEST_ASSETS_DIR}
```
siliconbrain, line 143: this directory should be added as a dependency with a matching target; see `bin`

Makefile
```diff
@@ -144,22 +137,14 @@ bin/golangci-lint_${GOLANGCI_VERSION}: | bin
...
+	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/$(ENVTEST_CTRL_VERSION)/hack/setup-envtest.sh
```
siliconbrain, line 144: this should also be a dependency and let make take care of testing whether the file is there

controllers/suite_test.go
```diff
@@ -111,19 +112,31 @@ func afterSuite() error {
 // duplicateRequest returns a reconcile.Reconcile implementation that delegates to inner and
 // writes the request to requests after Reconcile is finished.
 func duplicateRequest(t *testing.T, inner reconcile.Reconciler, stopped *bool) (reconcile.Reconciler, chan reconcile.Request, chan reconcile.Result, chan error) {
-	requests := make(chan reconcile.Request)
-	results := make(chan reconcile.Result)
-	errors := make(chan error)
+	requests := make(chan reconcile.Request, 100)
```
siliconbrain, line 115: why do we need buffered channels here? why with a capacity of 100?

Makefile
pepov: This example was pulled from a kubebuilder project with version v0.8.3, but kubebuilder itself hasn't been upgraded to v0.9.0 just yet. I expect there will be some other changes there as well, but this should be good enough for now.

Makefile
pepov: this target was mostly moved from an other project (kafka-operator) as is, and I expect it will change soon, but you are right in that this could be done much cleaner.

controllers/suite_test.go
pepov: I think you are right and we don't need buffered channels since we handle the case of the channels not beeing consumed already...

Makefile
siliconbrain: I mean there is better tooling for this in controller-runtime v0.9.0: https://github.com/kubernetes-sigs/controller-runtime/tree/v0.9.0/tools/setup-envtest

pepov: I saw this, I just didn't want to put more effort into this for now (oh dear laziness). But since we will use the same for the other projects we should use this I have to agree...

Makefile
pepov: I was referring to the new envtest tooling here just didn't find any examples, but given the above link we should be able to put it together ourselves.

Makefile
siliconbrain: Laziness is indeed a good reason not to do this 😄 but then we should create some issues as a reminder that this needs to be updated sometime in the future. Also, a comment marking this as a legacy solution would be nice so that others don't try to copy this as is if they can help it.

siliconbrain (approved):

PR #763, opened 2021-06-22
Abort draining on upscale

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Terminate the drainer job and remove the placeholder pod blocking the statefulset from creating its pods when it's upscaled.

### Why?
Without this, the statefulset is blocked from creating all its pods when upscaled until all drainer jobs have finished.

### DIFF
--- a/pkg/resources/fluentd/fluentd.go
+++ b/pkg/resources/fluentd/fluentd.go
@@ -16,6 +16,7 @@ package fluentd
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"emperror.dev/errors"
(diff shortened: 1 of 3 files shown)


pkg/resources/fluentd/fluentd.go
```diff
@@ -331,6 +336,21 @@ func (r *Reconciler) reconcileDrain(ctx context.Context) (*reconcile.Result, err
 			continue
 		}
 
+		if inUse && hasJob {
+			pvcLog.Info("deleting drainer job early as PVC is now in use")
+
+			if err := client.IgnoreNotFound(r.Client.Delete(ctx, &job, client.PropagationPolicy(v1.DeletePropagationBackground))); err != nil {
```
pepov, line 342: What would happen if we use foreground deletion here? Why background deletion is more appropriate? I have a sense that it would be better to keep the job there until the pod is completely removed to avoid any race conditions by accident...

siliconbrain: sorry, it was copy-paste 😅 changed it to force foreground deletion

pepov (approved):

ahma (approved):
LGTM Thanks @siliconbrain

PR #765, opened 2021-06-23
feat: allow passing environment vars to fluentd/bit pods

alejandroEsc:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Updates the API objects Fluentbit and Fluentd to contain additional environment variables to be passed to their pods.

### Why?

It can be quite necessary to set environment variables to our Fluentbit/Fluentd pods. These vars may be referenced in service's configurations, e.g., the hostname of the node where the pod may be running.

### Additional context

This is an API change but should be backwards compatible and not break existing deployments.

### Checklist


alejandroEsc:
it is unclear why the license check test is failing for the biuld, it is passing locally, any help with this?

ahma (approved):
Thanks @alejandroEsc LGTM

pepov (approved):

PR #766, opened 2021-06-23
feat: allow passing environment vars to nodeagent pods

alejandroEsc:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Updates the API object nodeagent to contain additional environment variables to be passed to its pods.

### Why?

This PR allows for environment variables to be passed to the nodeagent pods, allowing for potentially additional targeted configurations.

### Checklist


ahma (approved):
Thanks @alejandroEsc LGTM

pkg/sdk/api/v1beta1/node_agent_types.go
```diff
@@ -35,6 +36,7 @@ type NodeAgent struct {
 
 type NodeAgentFluentbit struct {
 	Enabled                 *bool                        `json:"enabled,omitempty"`
+	EnvVars                 []corev1.EnvVar              `json:"envVars,omitempty"`
```
pepov, line 39: This should already be doable using daemonset overrides, I wouldn't add this explicitly here.

To use this with daemonset overrides you just have to specify a subset of a daemonset config with just enough information to allow the merger find the path to the correct container. It should look like this:

```
daemonSet:
  spec:
    template:
      spec:
        containers:
        - name: <containerName>
          env: ...
```

pepov: In the logging CR
```
spec:
  nodeAgents:
  - name: <name>
    nodeAgentFluentbit:
      daemonSet:
        ...
```

alejandroEsc:
thank you for the info, closing for now.

PR #768, opened 2021-06-24
feat: create extraVolumes field for fluentd/bit and nodeagent

alejandroEsc:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This enables extraVolumes to be added to the fluentd/bit and nodeagent pods.

### Why?

In certain circumstances, e.g., audit logs in non-control-plane nodes, we would like to have volumes created in fluentbit so that it does not error on scraping a non-existing volume, in other cases we may need some data to remain persistent. 

### Checklist


ahma (approved):
Thanks @alejandroEsc

pepov:
I'm a bit hesitent since this will once again blow up the logging CRD size and we can't add fields indefinitely.

For the first use case I would say it could be cirumvented by defining different nodeagents for the different types of nodes. Correct me if that's not the case.

For the second use case, what data do you need to be persisted exactly?

alejandroEsc:
This would have been nice to have, but I will close for now, we are moving to our own fluentbit deployment for now. thank you.

pepov:
I would have been interested to understand the use case better, because I think it's not yet clear to me. Feel free to reach out over slack if you are open to discuss that further.

PR #771, opened 2021-06-28
fix: remove windows tolerations

mweibel:
when having default windows tolerations and setting tolerations in the
helm chart, the merge of those arrays don't work as expected.
not all clusters have taints on their windows nodes, in general
`nodeSelector` should work fine.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
Removes the tolerations from the windows nodeAgent daemonset to allow users specifying them as needed.

### Why?
When having predefined tolerations and trying to specify own tolerations the merge of the slice doesn't work as expected. 
E.g. I set the following via helm chart:
```
tolerations:
  - operator: Exists
```

To allow for any taint to be tolerated. The merge of the slices then ended up in the following:
```
tolerations:
  - key: "node.kubernetes.io/os"
    operator: Exists
    value: "windows"
    effect: NoSchedule
```
Which doesn't work and is actually an error (if `Exists` operator, no value can be set). 

### Checklist

### DIFF
--- a/pkg/resources/nodeagent/nodeagent.go
+++ b/pkg/resources/nodeagent/nodeagent.go
@@ -231,13 +231,6 @@ var NodeAgentFluentbitWindowsDefaults = &v1beta1.NodeAgent{
 						NodeSelector: map[string]string{
 							"kubernetes.io/os": "windows",
 						},
-						Tolerations: []v1.Toleration{{
-							Key:      "node.kubernetes.io/os",
-							Operator: "Equal",
-							Value:    "windows",
-							Effect:   "NoSchedule",
-						},
-						},
 					}},
 			}},
 	},


mweibel:
not sure what the description `Mergeable` check is about, TBH?

pepov:
Hey, the mergeable check was about the checkboxes and the items in the table I beleive, I fixed it.

I'm cool with the PR, however @nickgerace might have something to add, since they are using the windows node agent capabilities.

pepov (approved):

nickgerace:
Thanks @pepov. @mweibel, from what I am understanding, since there are default tolerations specified in the nodeagent object creation, they do not merge properly with conflicting specified in the Helm chart?

mweibel:
exactly. The merge seems to happen on index basis, i.e. if you have a list of each one item, they get merged together no matter if they are mergeable or not.

pepov:
the tolerations field in the podSpec does not define a `patchStrategy` and a `patchMergeKey`:
```
	Tolerations []Toleration `json:"tolerations,omitempty" protobuf:"bytes,22,opt,name=tolerations"`
```

With nodeAgents we switched to a method where we leverage the existing strategic merge patching capabilities of kubernetes, which works great in most situations, but not where the patch information is not defined or there are no unique keys, which is the case with tolerations I guess.

An example where this works well:
```
	Containers []Container `json:"containers" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,2,rep,name=containers"`
```

What you could do however in this case is to redefine the first toleration as is (or maybe an empty map would work as well) and define another one on top, but I agree, that in this case it's better to leave it off if possible.

nickgerace:
Ah I see, that makes sense from what I've been reading in the codebase as well @pepov. This LGTM as a result, but I do have one concern. @mweibel have you tested this with the default Windows taints and no additional tolerations? Just concerned about the "out of the box" experience.

mweibel:
The pods wouldn't go on windows nodes if they have a taint (which seems to be the case for GKE, f.e.). 
Personally I'm using a custom built cluster which doesn't have the taint. Rancher (f.e.) uses a different taint for the nodes IIRC and e.g. windows nodes on stock AKS clusters [aren't by default tainted](https://docs.microsoft.com/en-us/azure-stack/aks-hci/adapt-apps-mixed-os-clusters#taints-and-tolerations). 

I see your point about the default experience, however the taints seem to vary and therefore having a hard overridable default is bad. 
I see two possible resolutions:
1. make it a breaking change and document the behaviour change and the upgrade path (i.e. add the toleration if you have such taints)
2. add a new field to the windows nodeAgent (e.g. `addDefaultToleration: true`) and set it to true by default (which would add the node toleration) and with false it wouldn't add it. Isn't a pretty solution, though ;) 

Your thoughts?

pepov:
I'm ok with the first

ahma (approved):
lgtm Thanks @mweibel

nickgerace (approved):
Sounds fine to me with option 1 as well. We will track it for Rancher as well: https://github.com/rancher/rancher/issues/33572

mweibel:
:+1: thanks! 
Is there anything else I need to do in order to get this merged?

ahma:
@mweibel could you please update your branch from the master?

mweibel:
@ahma :+1: done.

PR #779, opened 2021-07-08
Feature/fluentd 1 13 test

ahma:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### Checklist


pepov (approved):

.github/workflows/docker_fluentd.yml
```diff
@@ -48,12 +48,17 @@ jobs:
         run: echo "::set-output name=value::$(echo ${TAG_OR_BRANCH##*/}| sed 's/fluentd-//g')"
         env:
           TAG_OR_BRANCH: ${{ steps.refortag.outputs.value }}
+      - name: Path from tag
+        id: pathfromtag
+        run: echo "::set-output name=value::$(echo ${TAG:0:5})"
```
siliconbrain, line 53:
```suggestion
        run: echo "::set-output name=value::$(echo $TAG | sed -E 's/^(v[0-9]+\.[0-9]+).*/\1/')"
```
would be more robust, it could work for versions `>1.*` too if we ever need that

siliconbrain (approved):
lgtm

PR #806, opened 2021-08-24
Error flow output

tarokkk:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

This PR add the feature to set ClusterOutput to the ERROR label. With this it is possible to grab events that caused error and send them to an Output


pkg/resources/model/system.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/banzaicloud/logging-operator/pkg/sdk/model/output"
+
```
siliconbrain, line 22:
```suggestion
```
sort imports

pkg/resources/model/system.go
```diff
@@ -159,6 +187,27 @@ func filtersForFilters(flowID string, flowName string, secretLoader secret.Secre
...
+			Tag:       "@ERROR",
+		},
+		FlowLabel: "@ERROR",
+	}
+
+	if clusterOutput := clusterOutputs.FindByName(outputRef); clusterOutput != nil {
+		plugin, err := plugins.CreateOutput(clusterOutput.Spec.OutputSpec, "main-fluentd-error", secrets.OutputSecretLoaderForNamespace(clusterOutput.Namespace))
+		if err != nil {
+			err = errors.WrapIff(err, "failed to create configured output %q", outputRef)
+			return nil, err
```
siliconbrain, line 203:
```suggestion
			return nil, errors.WrapIff(err, "failed to create configured output %q", outputRef)
```

siliconbrain (approved):

pepov (approved):

PR #809, opened 2021-08-31
add logging-operator specific PrometheusRule to resource builder

siliconbrain:
### What's in this PR?
Logging-operator specific Prometheus recording rules for Fluentd buffer metrics.

### Why?
These rules should provide metrics that are useful when monitoring Fluentd instances.


ahma (approved):
lgtm

pepov (approved):

PR #818, opened 2021-09-03
add missing RBAC rule for working with PrometheusRules

siliconbrain:
The operator cannot create PrometheusRules when enabled without this rule.


ahma (approved):
lgtm

pepov (approved):

PR #828, opened 2021-09-13
Makes log_group_name a required parameter in Cloudwatch output

fekete-robert:
A user has reported (see https://github.com/banzaicloud/logging-operator-docs/blob/e35a01cd72b1ece94f80aa855c38882db7f0068b/docs/configuration/plugins/outputs/cloudwatch.md ) that the Cloudwatch output doesn't work without this parameter, changing it to required in the source.


fekete-robert:
I'll need help regarding where/how to update the yamls so the build won't fail.

pepov:
I beleive not just log_group_name, but log_stream name is also required

pepov:
@fekete-robert you have to run `make check-diff` and it will highlight any changes that need to be added under version control

pepov (approved):

ahma (approved):
Thanks

phyrog:
@ahma I don't think this change was correct. There is no way to use `log_group_name_key` and `log_stream_name_key` now, as they are mutually exclusive to `log_group_name` and `log_stream_name`.

As you can see [here](https://github.com/fluent-plugins-nursery/fluent-plugin-cloudwatch-logs/blob/5b25fda45ee0eb06e85e76e3cad222854fd9df50/lib/fluent/plugin/out_cloudwatch_logs.rb#L84-L86), it is not allowed to set both fields, and `log_group_name` being required makes it thus impossible to use `log_group_name_key`. Same for `log_stream_name_key`.

ahma:
@phyrog Sorry but I missed your message, we'll coll back that https://github.com/banzaicloud/logging-operator/pull/1000

PR #834, opened 2021-09-21
Add scale subresource to logging object

velothump:
Allows HPA to scale fluentd replicas

| Q               | A
| --------------- | ---
| New feature?    | yes

### What's in this PR?
Adds scaling subresource to the logging object.

### Why?
Allows scaling fluentd via a HorizontalPodAutoscaler targeting the Logging object.

### Checklist


velothump:
Started this a few weeks ago, and now I've just seen https://github.com/banzaicloud/logging-operator/pull/826. Happy to close this down if it's preferable to scale on the statefulset directly instead.

pepov (commented):
Thanks for the PR! If that solution works for you as well, than yes it would be preferable.

PR #836, opened 2021-09-23
Reduce logging CRD size

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Removes v1beta1 logging spec and logging status from the v1alpha1 version of the logging resource.

### Why?
The v1alpha1 version is just a placeholder to allow the conversion webhook setup being tested in preparation for a future upgrade from v1beta1. Unfortunately the size of the logging resource reached the max annotation size limit with kubectl apply and k8s-objectmatcher working on the same resource.

### Additional context
A use case for this is when a CRD is part of a helm chart and thus managed by argocd, since it not just creates but also upgrades CRDs (it uses kubectl apply under the hood). If that same CRD is being managed by an operator that uses objectmatcher, the two managers will both use an annotation that hits the annotation size limit if the CRD is too big.


siliconbrain (approved):
lgtm

ahma (approved):
Thanks

PR #884, opened 2021-11-22
Support new multiline core introduced in fluentbit 1.8

### LINKED ISSUE #868: support Multiline Parsing in FluentbitSpec
Since the 1.8 release of fluentbit, multiline parsing is supported as a core functionality (see https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/multiline-parsing)
The new multiline parser has built in support for crio logs that are split into multiple partial messages (see https://github.com/fluent/fluent-bit/issues/1316).
Without proper concateration of the crio logs on the fluentbit end, exception detection on the fluentd end fails/incomplete.

What worked for us in a test environment is to enable multiline with cri parser (on fluentbit) and use detect_exceptions (on fluentd) to catch exceptions.

FluentbitSpec should support `multline.parser` keyword and built in parsers

Alternative could be to use concat filter, not sure.

davidmohar:
Signed-off-by: David Mohar <d.mohar@sportradar.com>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #868

### What's in this PR?
Support for using `multiline.parser` in fluentbit.

### Why?
Currently some container runtimes like containerd & cri can write longer log messages separated into multiple lines in the log file. A special flag is used to indicate if the message is partial (additional lines will follow) or full. You can handle this by using `Parser_Firstline` & `Parser_N` fields but with new multiline core support introduced in fluentbit 1.8 this can be achieved by only setting `multiline.parser` to the appropriate value and fluentbit handles this automatically. Multiple parser can be added which allow different node types to be scraped by the same fluentbit configuration. See more on https://docs.fluentbit.io/manual/pipeline/inputs/tail#multiline-support.

### Additional context
`fluentbit v1.8.9` is needed because of https://github.com/fluent/fluent-bit/pull/4197.

This was tested on one of our clusters using `logging-operator v3.15.0` and manually overriding fluentbit config using `customConfigSecret`. Only `multiline.parser` line was added to the tail INPUT. The only side-effect we noticed was that the message field changes from `message` to `log` after parsing.

The change seems straight-forward and simple but can be properly tested if needed as part of this PR.

### Checklist


siliconbrain (approved):
I don't *love* the obvious duplication, but lgtm

ahma (approved):
Thanks @davidmohar LGTM

davidmohar:
I've rebased to resolve the conflicts in crds

PR #905, opened 2022-01-13
Support for disable fluentd buffer

### LINKED ISSUE #901: Loki output buffers section mess
**Describe the bug**:
Since [this change](https://github.com/banzaicloud/logging-operator/pull/751) loki output has become very buggy. In my setup (using the mirrored rancher helm chart) the configuration was working fine with no buffers defined in the cluster output. Since this pull request, that has added default buffers for all outputs, the loki server has stopped receiving logs consistently. 
In the fluentd pod the buffer files are pilling up and at some time stop flushing altogether. Playing with buffer parameters solved nothing. The buffer files were not being flushed as desired. 
The point is - I don't want buffers at all. Or at least I would like memory buffers. However the type: memory buffer setting is not valid anymore, because the above pull request is adding Path parameter regardless of the type and the configcheck is producing a warning: 
```
[warn]: parameter 'path' in <buffer tag,time>
@type "memory"
...
path /buffers/..redacted__.*.buffer
...
</buffer> is not used. 
```
Which is some workaround, but in the generated app-config file the buffer section is still there. And there is some delay and inconsistency in the received logs by the loki server. 
Currently I have manually edited the secret, removing the <buffers> from the loki section and am praying for the logging operator to not overwrite it it the buggy one. 

**Expected behaviour**:

- When you want no buffers, defining no buffers in the output yaml should produce no such, or at least having a parameter like buffer.disabled or buffer.defaults, which would describe the behavior better. 
- Being able to use memory-type buffers when you want so. Currently it is impossible. 

Running the logging operator with version > 3.12 and a loki output with no buffers section or with buffers.type: memory parameter. 

**Environment details**:
- Kubernetes version (e.g. v1.15.2): v1.21.5
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): RKE
- logging-operator version (e.g. 2.1.1): 3.15
- Install method (e.g. helm or static manifests): [rancher mirrored helm chart](https://github.com/rancher/charts/tree/release-v2.6/charts/rancher-logging/rancher-logging/100.0.1%2Bup3.15.0) 
```
kind: ClusterOutput
metadata:...
loki:
    buffer:
      disable_chunk_backup: true
      flush_interval: 1s
      flush_mode: interval
      timekey: 1m
      timekey_wait: 5s
      type: memory
    extract_kubernetes_labels: false
    labels:...
```
/kind bug

ahma:
- memory buffer path fix

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #901

### Checklist


pkg/sdk/logging/model/render/fluent.go
```diff
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"emperror.dev/errors"
-
 	"github.com/banzaicloud/logging-operator/pkg/sdk/logging/maps/mapstrstr"
 	"github.com/banzaicloud/logging-operator/pkg/sdk/logging/model/types"
```
siliconbrain, line 25: These sections should stay separate.

siliconbrain (approved):
lgtm

PR #917, opened 2022-01-20
feat: add possibility to ref a secret in security field

cyril-corbon:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #X

### What's in this PR?

add a field to ref a secret for the security section

### Why?

to allow user to specify a secret 

### Checklist

### To Do

- [x] If the PR is not complete but you want to discuss the approach, list what remains to be done here


cyril-corbon:
should I change the type of SharedKey from `String` to `secret.Secret` ? 🤔 
it could break a lot configurations but I think it would be more logic
in the meanwhile I added another parameter ShareKeySecretRef as secret and set both of them as not **required**

ahma:
Thanks for the contribution. [fmi](https://community-banzaicloud.slack.com/archives/CGMPPBZBK/p1648208841914259?thread_ts=1647949183.959449&cid=CGMPPBZBK)

PR #935, opened 2022-03-01
Fluentbit modify filter plugin support

ahma:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### Checklist

- [ x] Related Helm chart(s) updated (if needed)


siliconbrain (approved):
lgtm

pkg/sdk/logging/api/v1beta1/fluentbit_types.go
```diff
@@ -278,6 +281,99 @@ type FilterAws struct {
 	Match string `json:"Match,omitempty" plugin:"default:*"`
 }
 
+// FilterModify The Modify Filter plugin allows you to change records using rules and conditions.
+type FilterModify struct {
```
siliconbrain, line 285: Maybe we should call this type `ModifyFilter`, then the plural form (`ModifyFilters`) would look better as a field name for the slice. But I know that's a lot of renaming and we might not want to put that much effort into this code base, so I'm fine either way.

siliconbrain (approved):

tarokkk (approved):

PR #969, opened 2022-04-03
allow for the configuration of the fluentd elasticsearch-genid filter…

staerion:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | N/A

### What's in this PR?
Add the option to configure the elasticsearch_genid filter from the banzai logging operator. This is a small extension to the current functionality that the banzai operator already offers to push loggings to elasticseearch.

### Why?
The [elasticsearch_genid plugin](https://github.com/uken/fluent-plugin-elasticsearch/blob/master/README.ElasticsearchGenID.md) is released together with the [fluentd elasticesarch plugin](https://github.com/uken/fluent-plugin-elasticsearch). It allows for generating hashes based on a document's properties. These unique hashes can then be used as the document id before sending it to elasticsearch. This provides a convenient and documented way to prevent elasticsearch from accepting the same document multiple times (in the case of partially rejected chunks for example. The convenience of this feature is documented in the fluent-plugin-elasticearch docs (here)[https://github.com/uken/fluent-plugin-elasticsearch#id_key]

### Additional context
Currently we can find many duplicates of log entries when fluentd tries to push a backlog of buffer loggings to elasticsearch. When chunks are (partially) rejected the whole chunk will be offered to fluentd again even when some of those documents in there have already been processed by elasticsearch. As a result we'll find the same log entry multiple times in elasticsearch. Setting the document id in fluentd prevents this problem. This functionality is already present in the banzai operator, but currently there is no good way to generate a unique hash to serve as the document id. This PR will include that functionality. The added functionality will come at a performance cost when activated, but is it fully optional.

### Checklist

- [ X] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [ X] User guide and development docs updated (if needed)
- [ X] Related Helm chart(s) updated (if needed)


ahma (approved):
lgtm thanks @staerion

bshifter (approved):

pkg/sdk/logging/model/filter/elasticsearch_genid.go
```diff
@@ -0,0 +1,78 @@
...
+	// You can specify to use record in events for hash generation seed. This parameter should be used with record_keys parameter in practice.
+	Use_record_as_seed bool `json:"use_record_as_seed,omitempty"`
+	// You can specify keys which are record in events for hash generation seed. This parameter should be used with use_record_as_seed parameter in practice.
+	Record_keys string `json:"record_keys,omitempty"`
+	// You can specify to use entire record in events for hash generation seed.
+	Use_entire_record bool `json:"use_entire_record,omitempty"`
+	// You can specify separator charactor to creating seed for hash generation.
+	Separator string `json:"separator,omitempty"`
+	// You can specify hash algorithm. Support algorithms md5, sha1, sha256, sha512. Default: sha1
+	Hash_type string `json:"hash_type,omitempty"`
```
siliconbrain, line 39: These field names could conform to the Go naming convention (Pascal case in this instance).

siliconbrain (approved):
lgtm

pkg/sdk/logging/model/filter/elasticsearch_genid.go
ahma: @staerion could you check it please

staerion: @ahma updated

ahma: @staerion please take a look at [this](https://www.w3schools.com/go/go_variable_naming_rules.php) thanks

PR #963, opened 2022-04-01
Add ability to set read_from_head parameter for fluentbit

fust:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | nos
| Deprecations?   | no
| Related tickets | N/A
| License         | Apache 2.0

### What's in this PR?
This PR enables setting the read_from_head parameter for FluentBit from the Logging CR.

### Why?
Newly discovered files are read from the tail. This can result in missing log messages.
This PR allows the user to set the read_from_head parameter, which makes FluentBit read from the head of these files instead of the tail, thus making sure no log messages get lost.
See also: [FluentBit tail parameters](https://docs.fluentbit.io/manual/pipeline/inputs/tail/)

The option is set to false by default so current behaviour is preserved.

### Checklist

- [X] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [X] Related Helm chart(s) updated (if needed)

PR #979, opened 2022-04-14
Add ability to set read_from_head parameter for fluentbit

fust:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Replaces #963 as that got a bit messy
This PR enables setting the read_from_head parameter for FluentBit from the Logging CR.

### Why?
Newly discovered files are read from the tail. This can result in missing log messages.
This PR allows the user to set the read_from_head parameter, which makes FluentBit read from the head of these files instead of the tail, thus making sure no log messages get lost.
See also: [FluentBit tail parameters](https://docs.fluentbit.io/manual/pipeline/inputs/tail/)

### Checklist

- [X] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [X] Related Helm chart(s) updated (if needed)


ahma (approved):
Thanks

bshifter (approved):

PR #998, opened 2022-05-04
Fix doc generator

fekete-robert:
- Fix source code so doc generator runs
- Fix for issue#899
- Minor fixes 

@ahma: I've added doc stubs to ```type Operation struct ``` in pkg/sdk/logging/api/v1beta1/fluentbit_types.go, but we should add a proper description


ahma (approved):
LGTM

PR #1014, opened 2022-05-20
Buffer without chunk key

### LINKED ISSUE #961: create buffer with no chunk key

Enabling buffer based on tags with timekey to a loki output results in no log delivery regardless of timekey settings and flush_mode: interval seems to have zero effect.  If it is possible to do this currently, I am not able to figure it out.

Currently I have disabled the buffer but this is not great.

I would like to have the ability to generate a `buffer` section that looks like

```
  <buffer>
    @type file
    flush_interval 5s
  </buffer>
```

ahma:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #961

### Checklist


pkg/sdk/logging/model/output/s3.go
```diff
@@ -242,7 +243,7 @@ func (c *S3OutputConfig) ToDirective(secretLoader secret.SecretLoader, id string
 		if c.Buffer == nil {
 			c.Buffer = new(Buffer)
 		}
-		c.Buffer.Tags = OneEyeTags
+		c.Buffer.Tags = &OneEyeTags
```
siliconbrain, line 246: I'd rather we use the `StringPointer` utility function from operator-tools to convert the `const` to a `*string`
```suggestion
		c.Buffer.Tags = utils.StringPointer(OneEyeTags)
```

siliconbrain (dismissed):
lgtm

pkg/sdk/logging/model/output/s3.go
ahma: Great Thanks!

bshifter (approved):

aslafy-z (approved):

PR #1031, opened 2022-06-08
fix: allow customizing cluster domain

### LINKED ISSUE #1021: Cluster name is hardcoded as cluster.local
**Describe the bug**:
Since PR [#894](https://github.com/banzaicloud/logging-operator/pull/894) the cluster name has been hardcoded as `cluster.local`, causing problems if this is not the actual name. 

**Expected behaviour**:
Cluster name should be configurable and not hardcoded.

aslafy-z:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1021

### What's in this PR?

Enable users to customize cluster domain

### Checklist

### To Do

- [x] If the PR is not complete but you want to discuss the approach, list what remains to be done here


charts/logging-demo/templates/secret_tls.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.loggingOperator.tls.enabled }}
 {{ $ca := genCA "svc-cat-ca" 3650 }}
-{{ $cn := printf "%s-%s.%s.svc.cluster.local" (include "logging-demo.name" .) "fluentd" .Release.Namespace }}
+{{ $cn := printf "%s-%s.%s.svc.%s" (include "logging-demo.name" .) "fluentd" .Release.Namespace .Values.clusterDomain }}
```
aslafy-z, line 3:
```suggestion
{{ $cn := printf "%s-%s.%s.svc.%s" (include "logging-demo.name" .) "fluentd" .Release.Namespace .Values.loggingOperator.clusterDomain }}
```

pkg/resources/nodeagent/nodeagent.go
```diff
@@ -199,7 +199,7 @@ func NodeAgentFluentbitDefaults(userDefined **v1beta1.NodeAgent) (*v1beta1.NodeA
 var NodeAgentFluentbitWindowsDefaults = &v1beta1.NodeAgent{
 	FluentbitSpec: &v1beta1.NodeAgentFluentbit{
 		FilterKubernetes: v1beta1.FilterKubernetes{
-			KubeURL:       "https://kubernetes.default.svc.cluster.local:443",
```
aslafy-z, line 202: Should this patch be preserved?

vzemtsov:
This fix work for fluentBit Secret-TLS.

But logging operator generate secret with FluentBit Config File, where in OutPut block:
```
[OUTPUT]
    Name          forward
    Match         *
    Host          logging-operator-logging-fluentd.logging.svc.**cluster.local**
    Port          24240
```
Need change this cluster.local

pkg/resources/nodeagent/nodeagent.go
aslafy-z: Or we can simply remove it everywhere like I proposed in https://github.com/banzaicloud/logging-operator/issues/1021#issuecomment-1195832840. WTYT @ahma @siliconbrain

siliconbrain:
I'm only vaguely familiar with how DNS resolution works when the cluster domain is not specified. I know it works in most cases, but I'm sure there are some edge cases e.g. maybe with a service mesh (e.g. Istio) or other multi-cluster setup.

Also, the trailing dot that we leave in the generated domain name if cluster domain is defined as `""` might be problematic.

charts/logging-demo/templates/secret_tls.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.loggingOperator.tls.enabled }}
 {{ $ca := genCA "svc-cat-ca" 3650 }}
-{{ $cn := printf "%s-%s.%s.svc.cluster.local" (include "logging-demo.name" .) "fluentd" .Release.Namespace }}
+{{ $cn := printf "%s-%s.%s.svc.%s" (include "logging-demo.name" .) "fluentd" .Release.Namespace .Values.loggingOperator.clusterDomain }}
```
aslafy-z, line 3:
```suggestion
{{- if .Values.loggingOperator.clusterDomain }}
{{ $cn := printf "%s-%s.%s.svc.%s" (include "logging-demo.name" .) "fluentd" .Release.Namespace .Values.loggingOperator.clusterDomain }}
{{- else }}
{{ $cn := printf "%s-%s.%s.svc" (include "logging-demo.name" .) "fluentd" .Release.Namespace }}
{{- end }}
```

charts/logging-operator-logging/templates/secret.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.tls.enabled }}
 {{ $ca := genCA "svc-cat-ca" 3650 }}
-{{ $cn := printf "%s-%s.%s.svc.cluster.local" (include "logging-operator-logging.name" .) "fluentd" .Release.Namespace }}
+{{ $cn := printf "%s-%s.%s.svc.%s" (include "logging-operator-logging.name" .) "fluentd" .Release.Namespace .Values.clusterDomain }}
```
aslafy-z, line 3:
```suggestion
{{- if .Values.loggingOperator.clusterDomain }}
{{ $cn := printf "%s-%s.%s.svc.%s" (include "logging-operator-logging.name" .) "fluentd" .Release.Namespace .Values.clusterDomain }}
{{- else }}
{{ $cn := printf "%s-%s.%s.svc" (include "logging-operator-logging.name" .) "fluentd" .Release.Namespace }}
{{- end }}
```

charts/logging-operator-logging/values.yaml
```diff
@@ -68,6 +69,9 @@ skipInvalidResources: false
 
 # Limit namespaces from where to read Flow and Output specs
 watchNamespaces: []
+
+# Cluster domain name
+clusterDomain: "cluster.local"
 # Control namespace that contains ClusterOutput and ClusterFlow resources
```
aslafy-z, line 75:
```suggestion

# Control namespace that contains ClusterOutput and ClusterFlow resources
```

ahma (approved):
Looks fine to me, Thanks @aslafy-z

pkg/resources/fluentbit/configsecret.go
```diff
@@ -126,7 +126,7 @@ func (r *Reconciler) configSecret() (runtime.Object, reconciler.DesiredState, er
 
 	var fluentbitTargetHost string
 	if r.Logging.Spec.FluentdSpec != nil && r.Logging.Spec.FluentbitSpec.TargetHost == "" {
-		fluentbitTargetHost = fmt.Sprintf("%s.%s.svc.cluster.local", r.Logging.QualifiedName(fluentd.ServiceName), r.Logging.Spec.ControlNamespace)
+		fluentbitTargetHost = fmt.Sprintf("%s.%s.svc.%s", r.Logging.QualifiedName(fluentd.ServiceName), r.Logging.Spec.ControlNamespace, r.Logging.Spec.ClusterDomain)
```
siliconbrain, line 129: I think this format string will have the same problem with the trailing dot if `ClusterDomain` is set to empty as I noted earlier with the chart template.

pkg/resources/fluentbit/configsecret.go
```diff
@@ -360,10 +360,11 @@ func (r *Reconciler) generateUpstreamNode(index int32) upstreamNode {
 	podName := r.Logging.QualifiedName(fmt.Sprintf("%s-%d", fluentd.ComponentFluentd, index))
 	return upstreamNode{
 		Name: podName,
-		Host: fmt.Sprintf("%s.%s.%s.svc.cluster.local",
+		Host: fmt.Sprintf("%s.%s.%s.svc.%s",
```
siliconbrain, line 363: I think this format string will have the same problem with the trailing dot if `ClusterDomain` is set to empty as I noted earlier with the chart template.

pkg/resources/nodeagent/configsecret.go
```diff
@@ -184,7 +184,7 @@ func (n *nodeAgentInstance) configSecret() (runtime.Object, reconciler.DesiredSt
 			SharedKey: n.nodeAgent.FluentbitSpec.TLS.SharedKey,
 		},
 		Monitor:                 monitor,
-		TargetHost:              fmt.Sprintf("%s.%s.svc.cluster.local", n.FluentdQualifiedName(fluentd.ServiceName), n.logging.Spec.ControlNamespace),
+		TargetHost:              fmt.Sprintf("%s.%s.svc.%s", n.FluentdQualifiedName(fluentd.ServiceName), n.logging.Spec.ControlNamespace, n.logging.Spec.ClusterDomain),
```
siliconbrain, line 187: I think this format string will have the same problem with the trailing dot if `ClusterDomain` is set to empty as I noted earlier with the chart template.

pkg/resources/nodeagent/configsecret.go
```diff
@@ -317,10 +317,11 @@ func (n *nodeAgentInstance) generateUpstreamNode(index int32) upstreamNode {
 	podName := n.FluentdQualifiedName(fmt.Sprintf("%s-%d", fluentd.ComponentFluentd, index))
 	return upstreamNode{
 		Name: podName,
-		Host: fmt.Sprintf("%s.%s.%s.svc.cluster.local",
+		Host: fmt.Sprintf("%s.%s.%s.svc.%s",
```
siliconbrain, line 320: I think this format string will have the same problem with the trailing dot if `ClusterDomain` is set to empty as I noted earlier with the chart template.

pkg/resources/nodeagent/nodeagent.go
```diff
@@ -304,6 +304,10 @@ func (r *Reconciler) Reconcile() (*reconcile.Result, error) {
 			if err != nil {
 				return nil, err
 			}
+
+			// Overwrite Kubernetes endpoint with a ClusterDomain templated value.
+			NodeAgentFluentbitDefaults.FluentbitSpec.FilterKubernetes.KubeURL = fmt.Sprintf("https://kubernetes.default.svc.%s:443", r.Logging.Spec.ClusterDomain)
```
siliconbrain, line 309: I think this format string will have the same problem with the trailing dot if `ClusterDomain` is set to empty as I noted earlier with the chart template.

pkg/sdk/logging/api/v1beta1/logging_types.go
```diff
@@ -61,6 +61,9 @@ type LoggingSpec struct {
 	GlobalFilters []Filter `json:"globalFilters,omitempty"`
 	// Limit namespaces to watch Flow and Output custom resources.
 	WatchNamespaces []string `json:"watchNamespaces,omitempty"`
+	// Cluster domain name to be used when templating URLs to services (default: "cluster.local").
+	// +kubebuilder:validation:Optional
+	ClusterDomain string `json:"clusterDomain"`
```
siliconbrain, line 66: I think we should go with an optional string here to allow explicitly setting it to empty. This way, the user can force resolution of the "clusterless" domain name.
```suggestion
	ClusterDomain *string `json:"clusterDomain,omitempty"`
```

pkg/resources/fluentbit/configsecret.go
aslafy-z: I imagined the field could not be set to empty, either it takes the specified value, or it uses the defaulted value from https://github.com/banzaicloud/logging-operator/pull/1031/files#diff-911a398d9218f00dc5749daedc4d64b64f16f2f8b010c898dc28393b3454fd0dR143

WDYT?

siliconbrain: By leaving off the tail of the FQDN the user can have more flexibility in how domain queries are resolved, utilizing the `search` option of `resolv.conf` (e.g. through `spec.dnsConfig.searches` field of the Pod resource). I think, it would be a shame not to support this given how little effort it would take on our side. Also, I see no downside in supporting this, but I might be wrong. Do you have any arguments against?

aslafy-z: Looks good to me. Will work on this.

charts/logging-operator-logging/values.yaml
```diff
@@ -12,6 +12,7 @@ flowConfigOverride: ""
 nameOverride: ""
 fullnameOverride: ""
 
+
```
aslafy-z, line 15:
```suggestion
```

aslafy-z:
@siliconbrain I think I addresses all the comments you left. Please have another check.

ahma (approved):
LGTM

pkg/sdk/logging/api/v1beta1/logging_types.go
```diff
@@ -561,6 +563,16 @@ func (l *Logging) QualifiedName(name string) string {
 	return fmt.Sprintf("%s-%s", l.Name, name)
 }
 
+// ClusterDomainAsSuffix formats the cluster domain as a suffix, e.g.:
+// .Spec.ClusterDomain == "", returns ""
+// .Spec.ClusterDomain == "cluster.local", returns ".cluster.local"
+func (l *Logging) ClusterDomainAsSuffix() string {
+	if l.Spec.ClusterDomain == nil || *l.Spec.ClusterDomain == "" {
```
siliconbrain, line 570: Explicitly setting `ClusterDomain` to `""` should result in no suffix—as you wrote, but leaving it unset should result in the "default" `.cluster.local` suffix so that we don't introduce a breaking change. (Some people might rely on this suffix being set by default.)

siliconbrain (approved):
👍

ahma (approved):

PR #1072, opened 2022-08-16
Feature/syslogng support minor doc fixes

fekete-robert:
Adds syslog-ng plugins to the generated docs and minor fixes


ahma (approved):
lgtm thanks @fekete-robert

PR #1098, opened 2022-09-14
helm chart update

bshifter:
- container/env
- container/volumemounts
- podspec/volume added to helm charts template and values file
+ some automated code formatting

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Helm chart users were unable to enable webhooks and mount the necessary cert files. Helm chart updated, this way there is a possibility to add these options.

### Why?
Logging-extensions were ported into the operator, but some of the features wasn't accessable through helm charts (i.e.: tailer-webhook)

### Checklist


siliconbrain (commented):
I think this is 👍 but we should definitely separate the damage done by go 1.19's gofmt from the useful work in this PR and fix up the butchered doc comments (and license headers) before merging.
As an alternative, roll back all the changes done by gofmt and commit only the main change set. That should pass all checks and leave comment fixing for another day.

siliconbrain (approved):

ahma (approved):
lgtm

PR #1111, opened 2022-10-04
fluentd serviceMonitor - support aggregated_metrics scrape path

### LINKED ISSUE #1110: Multiworker support for aggregated_metrics endpoint in servicemonitor
**Describe the bug**:

Not sure if this is a bug or feature request since official support for multi workers is unclear. But since the fluentdSpec supports specifying workers count, I'm proceeding with a bug report.

When using multi-workers, metrics are now generated per worker process with a new label, worker_id. The metric endpoint is actually incremented by worker_id above worker0's port, which is the value defined in the prometheus source.

There is an [aggregated endpoint](https://github.com/fluent/fluent-plugin-prometheus/blob/1f67114fbbae619f9ccc422b1d558efc3d658150/README.md#prometheus-input-plugin) at default path `/aggregated_metrics` that can be scraped by prometheus for all workers.

Unfortunately, there doesn't seem to be a way to override the serviceMonitor's config for metrics endpoint since it uses the same variable as the metrics path, which we don't want to change.

**Expected behaviour**:
The servicemonitor object should see if there are more than one worker and use the aggregated metrics endpoint instead.

1. Set workers to > 1
2. Notice that prometheus metrics only have timeseries for worker_id = 0 (first worker)

**Additional context**:

**Environment details**:
- Kubernetes version (e.g. v1.15.2): v1.21
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): Rancher + EKS
- logging-operator version (e.g. 2.1.1): 3.17.4
- Install method (e.g. helm or static manifests): Helm
- Logs from the misbehaving component (and any other relevant logs): N/A

Logging object:
```yaml
kind: Logging
metadata:
  labels:
    app.kubernetes.io/managed-by: Helm
  name: banzaiLogging
spec:
  fluentbit:
  #<...snip...>
  fluentd:
    livenessDefaultCheck: true
    logLevel: warn
    metrics:
      path: /metrics
      port: 24231
      serviceMonitor: true
    resources:
      limits:
        memory: 1000Mi
      requests:
        cpu: "1"
        memory: 512Mi
    rootDir: /buffers
    scaling:
      replicas: 10
    security:
      roleBasedAccessControlCreate: true
    workers: 2
```

Resultant serviceMonitor object:
```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
#<...snip...>
spec:
  endpoints:
  - bearerTokenSecret:
      key: ""
    interval: 15s
    path: /metrics
    port: http-metrics
    scrapeTimeout: 5s
  namespaceSelector:
    matchNames:
    - namespace_name
  selector:
    matchLabels:
      app.kubernetes.io/component: fluentd
      app.kubernetes.io/name: fluentd
```

/kind bug

buddyledungarees:
https://github.com/banzaicloud/logging-operator/issues/1110

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1110 

### What's in this PR?
Change prometheus serviceMonitor scrape path if number of workers is greater than one.

### Why?
See #1110 

### Additional context
I debated putting it behind a flag but I can't think of a scenario where anyone using multi-workers wouldn't want to get metrics for all worker processes.

### Checklist
- [x ] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)

This is my quick, hacky attempt to fix #1110, feel free to trash or close if better approach desired.

### DIFF
--- a/pkg/resources/fluentd/service.go
+++ b/pkg/resources/fluentd/service.go
@@ -91,6 +91,10 @@ func (r *Reconciler) monitorServiceMetrics() (runtime.Object, reconciler.Desired
 				objectMetadata.Labels[k] = v
 			}
 		}
+		metricScrapePath := r.Logging.Spec.FluentdSpec.Metrics.Path
+		if r.Logging.Spec.FluentdSpec.Workers > 1 {
+			metricScrapePath = "/aggregated_metrics"
+		}
 
 		return &v1.ServiceMonitor{
 			ObjectMeta: objectMetadata,
@@ -100,7 +104,7 @@ func (r *Reconciler) monitorServiceMetrics() (runtime.Object, reconciler.Desired
 				PodTargetLabels: nil,
 				Endpoints: []v1.Endpoint{{
 					Port:                 "http-metrics",
-					Path:                 r.Logging.Spec.FluentdSpec.Metrics.Path,
+					Path:                 metricScrapePath,
 					Interval:             r.Logging.Spec.FluentdSpec.Metrics.Interval,
 					ScrapeTimeout:        r.Logging.Spec.FluentdSpec.Metrics.Timeout,
 					HonorLabels:          r.Logging.Spec.FluentdSpec.Metrics.ServiceMonitorConfig.HonorLabels,


aslafy-z (approved):

pkg/resources/fluentd/service.go
```diff
@@ -91,6 +91,10 @@ func (r *Reconciler) monitorServiceMetrics() (runtime.Object, reconciler.Desired
 				objectMetadata.Labels[k] = v
 			}
 		}
+		metricScrapePath := r.Logging.Spec.FluentdSpec.Metrics.Path
+		if r.Logging.Spec.FluentdSpec.Workers > 1 {
+			metricScrapePath = "/aggregated_metrics"
```
siliconbrain, line 96: If you override the metrics path here, users won't be able to customize the path (if they are using multiple workers). Instead, the default value should be set depending on the number of workers here: https://github.com/banzaicloud/logging-operator/blob/6b16d314e8f0da31ee292ab930150467af11bdd3/pkg/sdk/logging/api/v1beta1/logging_types.go#L181

pepov:
@buddyledungarees may I ask what was the reason that you closed this?

PR #1163, opened 2023-01-20
Added copy output

petergvizd:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Added copy output according to https://docs.fluentd.org/output/copy.

### Why?

To be able to copy events to multiple outputs.

### Checklist


Sebor (approved):

pepov:
@petergvizd You can already do this by creating multiple flows with the same selectors, but referencing different outputs. Doesn't it accomplish, what you want to achieve?

gvizdpeter:
@pepov not really. Image use case when you want to export prometheus metrics based on number for all records per generated fields and also output these records to Elasticsearch. According to your suggestion if I have 100 flow for ELS output I would need to create another 100 flows for prometheus, which looks suboptimal to me.

PR #1168, opened 2023-01-22
add option for loki output to configure "include_thread_label"

eljohnson92:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets |  

### What's in this PR?

This exposes the `include_thread_label` option added in 1.2.19 of the fluent-loki [plugin](https://grafana.com/docs/loki/latest/clients/fluentd/#output-format) 

### Why?

When using more than 1 flush_thread_count the associated thread is added as a label to each log line. This is unnecessary in some use cases and increases the cardinality of labels.

### Additional context

This requires 1.2.19 of the [fluent-plugin-grafana-loki](https://rubygems.org/gems/fluent-plugin-grafana-loki/versions/1.2.19) and the current image `ghcr.io/banzaicloud/fluentd:v1.14.6-alpine-6` only has version 1.2.18. It wasn't clear to me how this image gets updated, but since the gem version is not pinned in the [Dockerfile](https://github.com/banzaicloud/logging-operator/blob/master/fluentd-image/v1.14/Dockerfile#LL57C14-L57C14) I believe a new build of the image should pull in this new plugin version. Let me know if there is something additional that needs to be done as part of this PR.

### Checklist


ahma (approved):
LGTM Thanks @eljohnson92

PR #1174, opened 2023-01-31
Move persist file next to buffers

tarokkk:
Syslog-ng stores persist information (like buffer file path) in a persist file. In the case this file is not persisted Pod restart may spawn new buffer files and ignore old files. To prevent this we moved the persist file next to the buffers.

### DIFF
--- a/pkg/resources/syslogng/statefulset.go
+++ b/pkg/resources/syslogng/statefulset.go
@@ -89,7 +89,7 @@ func syslogNGContainer(spec *v1beta1.SyslogNGSpec) corev1.Container {
 		Args: []string{
 			"--cfgfile=" + configDir + "/" + configKey,
 			"--control=" + socketPath,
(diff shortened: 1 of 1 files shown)


ahma (approved):
lgtm

pkg/resources/syslogng/statefulset.go
```diff
@@ -89,6 +89,7 @@ func syslogNGContainer(spec *v1beta1.SyslogNGSpec) corev1.Container {
 		Args: []string{
 			"--cfgfile=" + configDir + "/" + configKey,
 			"--control=" + socketPath,
+			"--persist-file=" + bufferPath + "/syslog-ng.persist",
```
siliconbrain, line 92: use `filepath.Join` to avoid double slash
```suggestion
			"--persist-file=" + filepath.Join(bufferPath, "/syslog-ng.persist"),
```

siliconbrain (approved):

ahma (approved):

PR #1200, opened 2023-02-14
[feature] mattermost plugin

Sebor:
This PR adds support of fluentd mattermost [plugin](https://github.com/levigo-systems/fluent-plugin-mattermost)


pkg/sdk/logging/model/output/mattermost.go
```diff
@@ -0,0 +1,86 @@
...
+// +version:"0.2.2"
+// +description:"Sends logs to Mattermost via webhooks."
+// +status:"GA"
+type _metaMattermost interface{} //nolint:deadcode,unused
+
+// +kubebuilder:object:generate=true
+// +docName:"Output Config"
+type MattermostOutputConfig struct {
+	// webhook_url Incoming Webhook URI (Required for Incoming Webhook mode).
+	WebhookURL string `json:"webhook_url"`
```
aslafy-z, line 55: Shouldn't this value be extracted from a secret?

Sebor: You are right. Fixed

Sebor:
There are strange errors in CI e2e test:
```
# github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1alpha1
Error: pkg/sdk/logging/api/v1alpha1/output_conversion.go:28:32: cannot convert o.Spec (variable of type OutputSpec) to type v1beta1.OutputSpec
Error: pkg/sdk/logging/api/v1alpha1/output_conversion.go:40:22: cannot convert src.Spec (variable of type v1beta1.OutputSpec) to type OutputSpec
```
But I did not change those files

Sebor:
HI @ahma, @tarokkk !
Could you review my PR?

ahma:
@Sebor could you please update this branch?

pkg/sdk/logging/model/output/mattermost.go
```diff
@@ -0,0 +1,86 @@
...
+	// channel_id the id of the channel where you want to receive the information.
+	ChannelID string `json:"channel_id,omitempty"`
+	// message_color color of the message you are sending, the format is hex code. (default: #A9A9A9)
+	MessageColor string `json:"message_color,omitempty"`
+	// message_title title you want to add to the message. (default: fluent_title_default)
+	MessageTitle string `json:"message_title,omitempty"`
+	// message The message you want to send, can be a static message, which you add at this point, or you can receive the fluent infos with the %s
+	Message string `json:"message,omitempty"`
+	// enable_tls you can set the communication channel if it uses tls. (default: true)
+	EnableTLS bool `json:"enable_tls,omitempty"`
```
pepov, line 65: This should be declared as a pointer, otherwise it cannot be set to false when needed.

You can add a test to validate the expected behaviour for both use cases (implicit default: true, explcit: false)

pkg/sdk/logging/model/output/mattermost.go
```diff
@@ -0,0 +1,86 @@
...
+	// webhook_url Incoming Webhook URI (Required for Incoming Webhook mode).
+	WebhookURL *secret.Secret `json:"webhook_url"`
+	// channel_id the id of the channel where you want to receive the information.
+	ChannelID string `json:"channel_id,omitempty"`
+	// message_color color of the message you are sending, the format is hex code. (default: #A9A9A9)
+	MessageColor string `json:"message_color,omitempty"`
+	// message_title title you want to add to the message. (default: fluent_title_default)
+	MessageTitle string `json:"message_title,omitempty"`
+	// message The message you want to send, can be a static message, which you add at this point, or you can receive the fluent infos with the %s
+	Message string `json:"message,omitempty"`
```
pepov, line 63: In case we ever want to set any of these to the empty string "" we have to declare these as pointers.

pepov (changes requested):

pkg/sdk/logging/model/output/mattermost.go
```diff
@@ -0,0 +1,86 @@
...
+// +docName:"Output Config"
+type MattermostOutputConfig struct {
+	// webhook_url Incoming Webhook URI (Required for Incoming Webhook mode).
+	WebhookURL *secret.Secret `json:"webhook_url"`
+	// channel_id the id of the channel where you want to receive the information.
+	ChannelID string `json:"channel_id,omitempty"`
+	// message_color color of the message you are sending, the format is hex code. (default: #A9A9A9)
+	MessageColor string `json:"message_color,omitempty"`
+	// message_title title you want to add to the message. (default: fluent_title_default)
+	MessageTitle string `json:"message_title,omitempty"`
```
pepov, line 61: in case we want to set any of these as the empty string "" we have to declare these as pointers

pkg/sdk/logging/model/output/mattermost.go
Sebor: You are right. Fixed

pkg/sdk/logging/model/output/mattermost.go
Sebor: I don't understand in which cases we want to set those fields as "" instead of just omit ones?

siliconbrain (dismissed):
lgtm

pkg/sdk/logging/model/output/mattermost.go
pepov: If you omit, then there will be some default title set (fluent_title_default). If you want to clear the title, you won't be able to. Not that you have to do it, I'm just pointing it out, that this way we won't be able to do that.

pepov:
could you please run a fresh `make generate manifests`?

Sebor:
> could you please run a fresh `make generate manifests`?

done

pepov (approved):

pepov:
I've added some changes to fix the errors and unblock this PR

Sebor:
> @Sebor once we pull this in please add a pr to include this into one of the docker images (the one in this repo is only a test image)

Is this [PR](https://github.com/kube-logging/fluentd-images/pull/17) enough?

pepov:
> > @Sebor once we pull this in please add a pr to include this into one of the docker images (the one in this repo is only a test image)
> 
> Is this [PR](https://github.com/kube-logging/fluentd-images/pull/17) enough?

I would also add it to the v1.15 image as well

ahma (approved):
Thanks @Sebor LGTM

Sebor:
Looks like it's already [there](https://github.com/kube-logging/fluentd-images/blob/main/v1.15/Dockerfile#L83)

pepov (approved):

PR #1213, opened 2023-02-23
replace fixed syslog-ng persist file flag with guessed dynamic

siliconbrain:
This fixes an issue when persist file creation fails because there is no `/buffers` directory in the syslog-ng statefulset's syslog-ng container.

The new logic removes the hard coded `--persist-file` flag from syslog-ng's args and only adds it if it finds a volume mount "for buffers".


ahma (approved):
Thanks @siliconbrain

PR #1215, opened 2023-02-24
concat filter separator added default

karloramone:
In order to configure fluentd as in the docs: https://github.com/fluent-plugins-nursery/fluent-plugin-concat
the field separator shouldn't have omitempty tag, it should be included in the fluentd config. When we specify empty string like this `separator "" `, it is not included in the configuration at all.

example of usage:

```
 spec:

	filters:
	  - concat:
	      partial_key: "partial_message"
	      separator: ""
	      n_lines: 10
	selectors: {}
	localOutputRefs:
	  - demo-output

 ```
Fluentd Config Result:

 ```
 <filter **>

	@type concat
	@id test_concat
	key message
	n_lines 10
	partial_key partial_message
	separator ""

 </filter>
 ```


pepov:
I beleive this will be the correct solution. I had to make some changes to allow an omitempty string to be set as an empty string when it is declared as a string pointer:
https://github.com/kube-logging/logging-operator/pull/1235

The solution in this PR would cause backward compatibility issues as it would set the separator to be empty in all cases where the user did not set it, which is wrong since the default value was "\n" in that case, so it would break their config.

PR #1235, opened 2023-03-08
Support optional empty strings and fix the separator field in the concat secret

### LINKED ISSUE #1184: Fluentd Concat filter wrong option on field
**Describe the bug**:
In [fluentd concat](https://github.com/fluent-plugins-nursery/fluent-plugin-concat) filter documentation and also in the operator [code](https://github.com/kube-logging/logging-operator/blob/master/pkg/sdk/logging/model/filter/concat.go) is described to use field `separator` as an empty string. The problem is the field separator in `pkg/sdk/logging/model/filter/concat.go` has an option `omitemoty` which doesn't allow us that configuration, parser removes the field completely. 

https://github.com/kube-logging/logging-operator/blob/master/pkg/sdk/logging/model/filter/concat.go#L43

`Separator string `json:"separator,omitempty"``

https://github.com/kube-logging/logging-operator/blob/master/pkg/sdk/logging/model/filter/concat.go#L93

```
// ## Example `Concat` filter configurations
// ```yaml
// apiVersion: logging.banzaicloud.io/v1beta1
// kind: Flow
// metadata:
//
//	name: demo-flow
//
// spec:
//
//	filters:
//	  - concat:
//	      partial_key: "partial_message"
//	      separator: ""
//	      n_lines: 10
//	selectors: {}
//	localOutputRefs:
//	  - demo-output
//
// ```
```

**Expected behaviour**:
Not to remove field `separator` while parsing.

**Environment details**:
- Kubernetes version v1.21.14

Please let me know if there is some workaround.

/kind bug

pepov:
- fix: handle intentionally empty strings properly when defined as a string pointer
- fix: concat separator field should have a setting where it is empty

fixes #1184


siliconbrain (approved):
lgtm

Issue #1240, opened 2023-03-09
fluentd filter GeoIP skip_adding_null_record option does not work (always true)


genofire:
> Bugs should be filed for issues encountered whilst operating logging-operator.
> You should first attempt to resolve your issues through the community support
> channels, e.g. Slack, in order to rule out individual configuration errors. [#logging-operator](https://slack.banzaicloud.io/)
> Please provide as much detail as possible.

**Describe the bug**:
golangs default value of an boolean is `false` and is always overwritten by default handler to `true`.

**Expected behaviour**:
th config file has written `false`

**Steps to reproduce the bug**:
see #1239 

**Additional context**:
i write an test here #1239 

**Environment details**:
- logging-operator version (e.g. 2.1.1): 4.0.0

/kind bug

PR #1239, opened 2023-03-09
fix(fluentd/geoip-filter): add test to verify if skip_adding_null_record works correct

genofire:
this PR adds a test, which shows the bug descripted in #1240.

PR #1241, opened 2023-03-09
 fix(fluentd/geoip-filter): use point of bool for skip_adding_null_record to let default work correct

genofire:
**untested code**

solved #1240 

based on #1239 which contains the test to show the bug (which is here solved).


genofire:
Your CLA bot is really annoying

genofire:
fix space in #1239 - so this PR should now become green (and #1239 still failed ....)

genofire:
Like progosed, CI is green :)

PR #1249, opened 2023-03-14
Introduce NodeAgent CRD

pepov:
Introduce NodeAgent CRD and update the code to properly handle it, while preserving compatibility with the previous inline nodeagent config in the Logging resource. The new resource does not have a `name` field in the spec as it has that in the metadata.

Conflicts between legacy and new nodeagents (when they have the same name) handled by logging an error but moving on with the values configured in the NodeAgent CR.

Charts have been restored under `e2e/charts` so that we can properly run the e2e tests. We should keep the charts here and synchronize to github.com/kube-logging/helm-charts before release.


ahma (approved):

tarokkk (dismissed):

controllers/logging/logging_controller.go
```diff
@@ -181,8 +182,31 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+
+	if len(logging.Spec.NodeAgents) > 0 || len(loggingResources.NodeAgents) > 0 {
+		// load agents from standalone NodeAgent resources and additionally with inline nodeAgents from the logging resource
+		// for compatibility reasons
+		agents := make(map[string]loggingv1beta1.NodeAgentConfig)
+		for _, a := range loggingResources.NodeAgents {
+			agents[a.Name] = a.Spec.NodeAgentConfig
+		}
+		for _, a := range logging.Spec.NodeAgents {
+			if !agentExists(a.Name, loggingResources.NodeAgents) {
```
siliconbrain, line 202:
```suggestion
			if _, exists := agents[a.Name]; !exists {
```

siliconbrain (dismissed):
lgtm

ahma (approved):

siliconbrain (approved):
lgtm

PR #1254, opened 2023-03-19
Syslog-ng Redis output

ahma:


pkg/sdk/logging/model/syslogng/output/redis.go
```diff
@@ -0,0 +1,70 @@
...
+type _hugoRedis interface{} //nolint:deadcode,unused
+
+// +docName:"Sending messages from a local network to the Redis server"
+//
+// ## Prerequisites
+//
+// ## Example
+//
+// {{< highlight yaml >}}
+// apiVersion: logging.banzaicloud.io/v1beta1
```
tarokkk, line 28: We should fix indentation

PR #1258, opened 2023-03-20
Archive docs trigger

pepov:


pepov:
because I expect we will use this, just in a different way

siliconbrain (approved):

PR #1281, opened 2023-03-29
chore: charts up

pepov:
And sync to https://github.com/kube-logging/helm-charts/pull/15

### DIFF
--- a/charts/logging-operator-logging/README.md
+++ b/charts/logging-operator-logging/README.md
@@ -1,3 +1,7 @@
 # Logging operator logging chart
 
-The chart has been relocated [here](https://github.com/kube-logging/helm-charts/tree/main/charts/logging-operator-logging).
+This chart has been relocated under e2e/charts where it is used to run the e2e tests with.
+
+All modifications should go there first, get tested and then synchronized to https://github.com/kube-logging/helm-charts/ where it is linted and released.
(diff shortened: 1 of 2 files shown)


genofire:
Why still fill https://github.com/kube-logging/helm-charts ? and not make there ci part of this repo?

genofire:
Please also update / explain the different urls in:
https://github.com/kube-logging/logging-operator/blob/master/charts/logging-operator/README.md
and
https://github.com/kube-logging/logging-operator/blob/master/charts/logging-operator-logging/README.md

or move `e2e/charts` to `charts`

pepov:
Thanks for the reminder, I've just added notes about this state. I'm not relaxed to move the charts back to their original place until we fix this far from ideal state.

pepov:
> Why still fill https://github.com/kube-logging/helm-charts ? and not make there ci part of this repo?

Because the decision was to use a static helm repo temporarily (to replace the original banzaicloud repo), which requires the  charts to be in the given git repo. We will move the charts back here and implement an OCI based release later, but that requires some more time which we don't have at the very moment.

pepov:
@genofire I've sent you a collaborator invite, so that I can add you on PRs you might be interested in (like this one) as a reviewer from now on

PR #1285, opened 2023-03-29
fix(e2e/charts): add logging-operator-logging syslogNG + errorOutputRef support

genofire:
add support for syslogNG and errorOutputRef

some cleanups:
- replace `if` with `with`
- replace `indent` to `nindent` (and nicer indent)
- reorder like in docs / crd-struct (and take comments from crd-struct to helm-chart)

### DIFF
--- a/e2e/charts/logging-operator-logging/templates/logging.yaml
+++ b/e2e/charts/logging-operator-logging/templates/logging.yaml
@@ -28,8 +28,6 @@ spec:
     {{- with .Values.fluentbit }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
-  {{- else }}
-  fluentbit: {}
   {{- end }}
   {{- if or .Values.tls.enabled .Values.fluentd }}
   fluentd:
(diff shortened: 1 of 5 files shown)


e2e/charts/logging-operator-logging/templates/logging.yaml
```diff
@@ -5,69 +5,76 @@ metadata:
...
     {{- end }}
-    {{- if .Values.fluentbit }}
-{{ toYaml .Values.fluentbit | indent 4 }}
+    {{- with .Values.fluentd }}
+    {{- toYaml . | nindent 4 }}
     {{- end }}
   {{- else }}
-  fluentbit: {}
+  fluentd: {}
+  {{- end }}
```
pepov, line 47: this is incorrect I beleive (and it was incorrect previously as well)

this way we cannot disable fluentd completely, although that is a totally valid scenario as well (when syslog-ng replaces fluentd)

I can't add a suggestion due to the sorting but it should be this:
```
  {{- else if .Values.fluentd }}
  fluentd: {}
  {{- end }}
```

pepov (approved):
Nice addition and a very nice cleanup! One minor comment regarding an issue we could fix while we are here (see inline comments) and another minor thing: next time could you please put the change, the reorder and the cleanup part into separate commits so that it is easier to review? Thanks a lot again!

pepov (approved):

PR #1291, opened 2023-04-05
feat(nodeagent)!:disable nodeagent crd processing by default

pepov:
- make NodeAgent CRD processing optional (disabled by default)
- replace legacy cached rest mapper with `apiutil.NewDynamicRESTMapper` which is supported upstream already


ahma (approved):
LGTM

bshifter (approved):

PR #1298, opened 2023-04-12
chore(fix): syslog-ng persist file argument escaping

pepov:
Specify the persist-file path in a separate argument in order to escape special characters correctly


pkg/resources/syslogng/statefulset.go
```diff
@@ -85,7 +85,7 @@ func (r *Reconciler) statefulset() (runtime.Object, reconciler.DesiredState, err
 	if mnt := kubetool.FindVolumeMountByName(syslogngContainer.VolumeMounts, buffersVolumeName); mnt != nil {
 		if !sliceAny(syslogngContainer.Args, func(arg string) bool { return strings.Contains(arg, "--persist-file") }) {
 			persistFilePath := filepath.Join(mnt.MountPath, "/syslog-ng.persist")
-			persistFileFlag := fmt.Sprintf("--persist-file=%q", persistFilePath)
+			persistFileFlag := fmt.Sprintf("--persist-file=%s", persistFilePath)
```
siliconbrain, line 88: If `persistFilePath` contains any space the arg will be split and the value set for the persist file flag will only contain the part before the space. Instead, use this syntax:
```suggestion
			persistFileFlag := fmt.Sprintf("--persist-file %q", persistFilePath)
```
This handles spaces in `persistFilePath` correctly.

siliconbrain (changes requested):

siliconbrain (approved):

PR #1303, opened 2023-04-14
feat: update cluster domain to force direct resolution

aslafy-z:
As we generate FQDNs with this clusterDomain value, we can directly append a dot at the end to reduce dns queries in non optimal DNS configurations.

TODO: update tests


aslafy-z:
Replaced by https://github.com/kube-logging/logging-operator/pull/1303

Issue #1057, opened 2022-07-21
Documentation for extraVolumes option


abierbaum:
**Is your feature request related to a problem? Please describe.**

I would like to see documentation or an example added showing how to use the `extraVolumes` configuration with the `Logging` CRD and fluentd.  

**Describe the solution you'd like**

An example of how to correctly configure it similar to what is shown here: https://banzaicloud.com/docs/one-eye/logging-operator/configuration/fluentd/  for `bufferStorageVolume`.

**Describe alternatives you've considered**

I searched the source code, issues, and example and didn't find a place this is documented.  Also looked through the CRD definition file.

**Additional context**

I tried creating it the best I could and came up with something like below.  But when attempting to use it can't start fluentd.  I thought about filing a bug, but this isn't really a bug.  It is just a nice feature that isn't clear how to configure, so I think a documentation or example would be good.

```
apiVersion: logging.banzaicloud.io/v1beta1
kind: Logging
metadata:
  name: logging-demo
  namespace: logging
spec:
  controlNamespace: logging
  fluentbit:
    tls:
      enabled: true
      secretName: logging-demo-fluentbit-tls
      sharedKey: ""
  fluentd:
    logLevel: debug
    tls:
      enabled: true
      secretName: logging-demo-fluentd-tls
      sharedKey: ""
    extraVolumes:
      - path: "/data"
        containerName: fluentd
        volume:
          pvc:
            spec:
              accessModes:
                - ReadWriteOnce
              resources:
                requests:
                  storage: 30Gi
              storageClassName: logging-storage
              volumeMode: Filesystem
```

aslafy-z:
An helpful example could be configuring fluentd to store discarded chunks (`/tmp/fluent/backup`) in a PVC instead of on the host (ephemeral-config).

abierbaum:
@aslafy-z Do you have an example of a working configuration for extraVolumes?  I never got past that point and instead started misusing the buffer volumes configuration instead.

aslafy-z:
Ow, so I misread your issue from the beginning. I thought this example was working on your side and you created this issue to track the addition of this extraVolumes key to the documentation. My bad :sweat_smile:

So I guess you hit the same issue as I reported in https://github.com/banzaicloud/logging-operator/issues/1063 isn't it? It's not working on my side either.

starbops:
[...]

Issue #1063, opened 2022-07-25
extraVolumes PVC provisioning does not work on fluentd statefulset


aslafy-z:
**Describe the bug**:
PVC ExtraVolumes are not well added on Fluentd statefulset.

**Expected behaviour**:

I want the ExtraVolume to add a PVC as a volumeClaimTemplate so I get one extra volume per fluentd replicas.

**Steps to reproduce the bug**:

- Deploy a default Logging with no customization:
- Update with the following extraVolumes definition:
```yaml
          fluentd:
            # Adding an extra volume for discarded chunks so they doesn't fills up the host disk
            extraVolumes:
              - volumeName: discarded-chunks
                path: "/tmp/fluent/backup"
                containerName: fluentd
                volume:
                  pvc:
                    spec:
                      accessModes:
                        - ReadWriteOnce
                      resources:
                        requests:
                          storage: 10Gi
                      storageClassName: standard
                      volumeMode: Filesystem
```
- STS is well updated with the volumeMounts section
  ```
          volumeMounts:
          - mountPath: /tmp/fluent/backup
          name: discarded-chunks
  ```
- STS is wrongly updated with the volume section
  ```
      volumes:
      - name: discarded-chunks
        persistentVolumeClaim:
          claimName: ""
  ```

**Additional context**:

**Environment details**:
- Kubernetes version (e.g. v1.15.2): v1.20.9
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): RKE
- logging-operator version (e.g. 2.1.1): 3.17.8
- Install method (e.g. helm or static manifests): Helm
- Logs from the misbehaving component (and any other relevant logs): 

```
[...]

PR #1304, opened 2023-04-17
fix: add missing logic to handle PVCs for extraVolumes

pepov:
Test this with:
```
make install
helm repo add kube-logging https://kube-logging.github.io/helm-charts
helm install logging-operator-logging kube-logging/logging-operator-logging
```

Use the following config for fluentd (storageClassName should be something supported on your provider, I've tested this on minikube):
```
  fluentd:
    extraVolumes:
    - containerName: fluentd
      path: /data
      volume:
        pvc:
          source:
            claimName: fluentd-data
          spec:
            accessModes:
            - ReadWriteOnce
            resources:
              requests:
                storage: 1Gi
            storageClassName: standard
            volumeMode: Filesystem
```

Then simply run the operator and see if the pvcs and the fluentd pod has been created properly
```
make run
```

Fixes: #1057 #1063


siliconbrain (approved):
lgtm

Issue #1099, opened 2022-09-15
too many open files


gricuk:
We're getting` too many open files `error in few clusters with image ghcr.io/banzaicloud/fluentd:v1.14.5-alpine-1 and higher.

It happens randomly and I can't provide any conditions, but could you please check the following PR https://github.com/fluent/fluentd/pull/3844. It was fixed  in [1.15.2 release](https://github.com/fluent/fluentd/releases/tag/v1.15.2).

Now we have downgraded fluentd to ghcr.io/banzaicloud/fluentd:v1.14.4-alpine-2 and all is okay with this version (but I'm not sure because as I mentioned above it happens randomly and not all clusters have this error)

**Environment details**:
- Kubernetes version v1.21.12+rke2r2:
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): RKE2
- logging-operator version (e.g. 2.1.1): logging-operator-3.17.9
- Install method (e.g. helm or static manifests): helm

/kind bug

1it:
It could be caused by `chunk_limit_size: 8MB`, which is set [here](https://github.com/banzaicloud/logging-operator/blob/0ca75c22d5a2b5c877603ad379be7486388e693c/pkg/sdk/logging/model/output/buffer.go#L47). This actually a nonsense, it should be at least 256MB (Fluentd default). 8MB is a default value for in memory buffer, not for file based.
I've had about 32k buffer files in one instance because of that and constant `too many open files` errors.

pepov:
According to the docs, the chunk_limit_size is now smarter as it uses 8MB for memory and 256MB for file buffers by default:
https://docs.fluentd.org/configuration/buffer-section#buffering-parameters

We can remove our default setting to rely on fluentd's settings.

giakinh0823:
This is my issue. I have 2 node in k8s. 
- Node1: master + worker: This is node error `too many open files` 
- Node2: worker: This is node running success with deamonset
Please help me. Thank you.

Here are my settings:
```
apiVersion: logging.banzaicloud.io/v1beta1
kind: Logging
metadata:
  name: ewallet-logging
  namespace: logging
spec:
  fluentd:
    image:
      pullPolicy: IfNotPresent
      repository: rancher/banzaicloud-fluentd
      tag: v1.11.5-alpine-9
    configReloaderImage:
      repository: jimmidyson/configmap-reload 
      tag: latest  # Lỗi thì thay đổi version
[...]

Issue #732, opened 2021-05-03
Default to file buffer for PVC based setup


pepov:
In case the user doesn't configure buffer and PVC is configured (which is the default) we should configure buffers by default to use filesystem buffering instead of letting fluentd defaulting to memory.

Extracted from: #730

PR #1313, opened 2023-04-23
docs on buffer

genofire:
i do not find the way the default values are setup.
but here is the documentation:
https://docs.fluentd.org/configuration/buffer-section

so maybe update that correctly based on then

### DIFF
--- a/pkg/sdk/logging/model/output/buffer.go
+++ b/pkg/sdk/logging/model/output/buffer.go
@@ -37,26 +37,30 @@ type _metaBuffer interface{} //nolint:deadcode,unused
 type Buffer struct {
 	// Disable buffer section (default: false)
 	Disabled bool `json:"disabled,omitempty" plugin:"default:false,hidden"`
-	// Fluentd core bundles memory and file plugins. 3rd party plugins are also available when installed.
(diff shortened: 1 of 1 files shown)


pepov:
Hey! Thanks you are right, that the definitions should reflect the actual state of the plugin, but there are a few things we have to consider:

1. Backwards compatibility
We need to consider if any of these changes mean a breaking change in existing deployments

2. Using the `plugin` tag for setting defaults
The `plugin:"default:XY"` tag should only be used in special situations. This is not very consistent throughout the code right now, but I would like to improve this. 
What the `plugin:default` tag does is it explicitly fills the value to something in the generated fluentd code. Now if that value is the same that fluentd would use if you would leave the field out, then what is the point of setting it?
The point of the `plugin:default` is to set values that are _different_ from what fluentd would set anyways. Of course we need to clearly articulate why this is needed.

In addition to this there are also other things we need to take care of, when making changes around buffer settings, see for example:
https://github.com/kube-logging/logging-operator/issues/1099
https://github.com/kube-logging/logging-operator/issues/732

I'm going to further comment on this PR soon, but for now I need to ask for your patience.

PR #1332, opened 2023-05-16
fluentbit multiagent

### LINKED ISSUE #1331: Multi-agent fluentbit
In case there are different node groupsin a cluster, operators might need to run fluentbit with different configurations on those different groups.

It wasn't possible until now, but with the introduction of a separate FluentbitAgent CRD we can already achive it.

In the first iteration the main goal is to have the exact same `fluentbitSpec` with the exact same logic that was available inside the `logging` resource, but now as a standalone type. 

One important compatibility guarantee we are aiming for is to enable a smooth migration path from moving fluentbit out from the logging resource. It should be as easy as creating a new `FluentbitAgent` with the same name as the `Logging` resource and simply move everything from Logging/spec.fluentbit to FluentbitAgent/spec. The only thing that should happen is that ownerrefs for managed resources will change from the Logging resource to the new FluentbitAgent

pepov:
Enabling fluentbit multiagent for a single logging resource and fix fluentbit-agent resource name
Make fluentbit-agent installed and processed by default
Extend e2e tests to print cluster logs from relevant namespaces to ease debugging

Fixes #1331


e2e/common/cluster.go
```diff
@@ -28,9 +34,13 @@ import (
 
 const defaultClusterName = "e2e-test"
 
+var NamespacesUsed []string
```
siliconbrain, line 37: Couldn't this be a field on the `Cluster` type, passed in as a parameter through `WithCluster`?

pepov: yes, I've changed the way it was done pretty much, see my last commit

e2e/volumedrain/volumedrain_test.go
```diff
@@ -334,6 +358,14 @@ func TestVolumeDrain_Downscale_DeleteVolume(t *testing.T) {
 		require.Eventually(t, cond.ResourceShouldBeAbsent(t, c.GetClient(), common.Resource(new(corev1.Pod), ns, fluentdReplicaName)), 30*time.Second, time.Second/2)
 
 		require.Eventually(t, cond.ResourceShouldBeAbsent(t, c.GetClient(), common.Resource(new(corev1.PersistentVolumeClaim), ns, logging.Name+"-fluentd-buffer-"+fluentdReplicaName)), 30*time.Second, time.Second/2)
+	}, func(t *testing.T, c common.Cluster) error {
+		path := fmt.Sprintf("%s/cluster-%s.log", TestTempDir, t.Name())
```
siliconbrain, line 362:
```suggestion
		path := filepath.Join(TestTempDir, fmt.Sprintf("cluster-%s.log", t.Name()))
```

e2e/common/cluster.go
```diff
@@ -78,7 +110,25 @@ func DeleteTestCluster(clusterName string) error {
...
+func (c kindCluster) PrintLogs(config PrintLogConfig) error {
+	cmd := exec.Command("stern", "-n", strings.Join(config.Namespaces, ","), ".*", "--no-follow", "--tail", cast.ToString(config.Limit), "--kubeconfig", c.kubeconfigFilePath)
+	f, err := os.Create(config.FilePath)
+	if err != nil {
+		return err
+	}
+	cmd.Stdout = f
+	cmd.Stderr = os.Stderr
+
+	defer f.Close()
```
siliconbrain, line 126:
```suggestion
	defer f.Close()

	cmd.Stdout = f
	cmd.Stderr = os.Stderr

```

siliconbrain (approved):

tarokkk (dismissed):

PR #1359, opened 2023-06-13
sync chart changes

pepov:
- fix: do not refer to upstream config item to avoid hiding fluentbit configuration, also use absolute path in the config
- chore: sync chart changes


ahma (approved):
lgtm

PR #1362, opened 2023-06-14
Feature syslog-ng Elasticsearch output

tarokkk:
This PR provides support for OpenSearch/Elasticsearch HTTP output vi syslog-ng. The goal is to be able to swap from fluentd elastic output (https://github.com/uken/fluent-plugin-elasticsearch) as smoothly as possible.

Options available:
- [x] All HTTP options
- [x] Logstash format (prefix, suffix) with good defaults
- [x] Index name
- [x] authentication (username, password)

How this was tested:
https://gist.github.com/OverOrion/57c918f2c75c35af30b2a0531924780d


OverOrion:
Seems like I can not ask for a review from the PR's author, so asking a review from @tarokkk here :).

OverOrion:
Also here are some notes regarding a simple test deployment for this branch:
https://gist.github.com/OverOrion/57c918f2c75c35af30b2a0531924780d

pkg/sdk/logging/model/syslogng/output/elasticsearch.go
```diff
@@ -0,0 +1,62 @@
...
+package output
+
+import "fmt"
+
+// +name:"Elasticsearch"
+// +weight:"200"
+type _hugoElasticsearch interface{} //nolint:deadcode,unused
+
+// +docName:"Sending messages over Elasticsearch"
+// More info at https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.37/administration-guide/destination-forward-send-and-store-log-messages/elasticsearch-http-sending-messages-to-elasticsearch-http-bulk-api/
```
pepov, line 24: would worth changing the urls to our docs: https://axoflow.com/docs/axosyslog-core/chapter-destinations/configuring-destinations-elasticsearch-http/

pkg/sdk/logging/model/syslogng/output/elasticsearch.go
```diff
@@ -0,0 +1,62 @@
...
+	if o.LogstashPrefixSeparator == "" {
+		o.LogstashPrefixSeparator = "-"
+	}
+
+	if o.LogStashSuffix == "" {
+		o.LogStashSuffix = "${YEAR}.${MONTH}.${DAY}"
+	}
+
+	if o.LogstashPrefix != "" {
+		o.Index = fmt.Sprintf("%s%s%s", o.LogstashPrefix, o.LogstashPrefixSeparator, o.LogStashSuffix)
```
pepov, line 60: what if the user wants to set the index explicitly?

pepov (approved):

pkg/sdk/logging/model/syslogng/output/elasticsearch.go
tarokkk: Then he should set the index field, and not set the Logstashprefix. This just helps if you don't want to bother with the formatting but want logstash format. What would be interesting to syslog-ng template in the prefix as well like pod-name.

OverOrion (approved):
Approving in @tarokkk's stead.

PR #1379, opened 2023-07-12
Move helm chart back to this repo

sagikazarmark:


.github/workflows/artifacts.yaml
```diff
@@ -1,71 +1,262 @@
...
           platforms: linux/amd64,linux/arm64,linux/arm/v7
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
-          push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          outputs: ${{ steps.build-output.outputs.value }}
+          # push: ${{ inputs.publish }}
```
pepov, line 115: Don't we need this?

.github/workflows/artifacts.yaml
```diff
@@ -1,71 +1,262 @@
...
+
+      - name: Extract OCI tarball
+        run: |
+          mkdir -p image
+          tar -xf image.tar -C image
+
+      # See https://github.com/anchore/syft/issues/1545
+      - name: Extract image from multi-arch image
+        run: skopeo --override-os linux --override-arch amd64 --insecure-policy copy --additional-tag ${{ steps.image-name.outputs.value }}:${{ steps.meta.outputs.version }} oci:image docker-archive:docker.tar
+
```
pepov, line 139:
```suggestion
        if: inputs.publish

```

.github/workflows/artifacts.yaml
```diff
@@ -1,71 +1,262 @@
...
+          outputs: ${{ steps.build-output.outputs.value }}
+          # push: ${{ inputs.publish }}
+
+      - name: Set image ref
+        id: image-ref
+        run: echo "value=${{ steps.image-name.outputs.value }}@${{ steps.build.outputs.digest }}" >> "$GITHUB_OUTPUT"
+
+      - name: Fetch image
+        run: skopeo --insecure-policy copy docker://${{ steps.image-name.outputs.value }}:${{ steps.meta.outputs.version }} oci-archive:image.tar
+        if: inputs.publish
```
pepov, line 123: this step seems to be the prerequisite of the subsequent steps, isn't this condition breaking them?

.github/workflows/artifacts.yaml
```diff
@@ -1,71 +1,262 @@
...
+          name: "[${{ github.job }}] Helm chart"
+          path: ${{ steps.build.outputs.package }}
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+        if: inputs.publish && inputs.release
```
pepov, line 230:
```suggestion
        if: inputs.publish
```

.github/workflows/artifacts.yaml
```diff
@@ -1,71 +1,262 @@
...
+
+      - name: Helm push
+        run: helm push ${{ steps.build.outputs.package }} oci://${{ steps.oci-registry-name.outputs.value }}
+        env:
+          HELM_REGISTRY_CONFIG: ~/.docker/config.json
+        if: inputs.publish && inputs.release
```
pepov, line 236:
```suggestion
        if: inputs.publish
```

Can we let helm charts from the master branch for example released under version 0.0.0-master.$buildNumber?

charts/logging-operator/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - fluentd
+  - fluentbit
+home: https://kube-logging.github.io
+sources:
+  - https://github.com/kube-logging/logging-operator
+  - https://github.com/kube-logging/helm-charts/tree/main/charts/logging-operator
+annotations:
+  artifacthub.io/changes: |
+    - kind: changed
+      description: Moved to to location
```
pepov, line 19: Something left off from here maybe?

charts/logging-operator/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+  artifacthub.io/images: |
+    - name: logging-operator
+      image: ghcr.io/kube-logging/logging-operator:4.2.1
```
pepov, line 22: I beleive we should leave it off to avoid getting out of sync with the appVersion all the time.

> By default, Artifact Hub will try to extract the containers images used by Helm charts from the manifests generated from a dry-run install using the default values. If you prefer, you can also provide a list of containers images manually by using this annotation. Providing the information manually also allows adding some extra information, like the platforms supported by each of the images.

```suggestion
```

.github/workflows/artifacts.yaml
sagikazarmark: Answered it offline

.github/workflows/artifacts.yaml
sagikazarmark: Answered it offline

.github/workflows/artifacts.yaml
sagikazarmark: The release bit is necessary for the moment, because publish is true for the default branch.

.github/workflows/artifacts.yaml
sagikazarmark: Currently you can tag 0.0.0-dev.X versions. What you are proposing is feasible...but let's do that in a subsequent PR.

pepov (approved):

PR #1383, opened 2023-07-19
fix append slice

eumel8:
If prometheus crd are available, resourceObjects get lost for reconciling. In effect no Statefulset for fluentd will created.

solved: https://github.com/kube-logging/logging-operator/issues/1384


pepov (approved):

ahma (approved):
Thanks @eumel8 LGTM

Issue #1386, opened 2023-07-20
Refuse to reconcile multiple logging resources with the same loggingRef


pepov:
**Describe the bug**:
The behaviour for the logging operator is unpredictable if there are multiple logging resources using the same loggingRef. (A logging resource without an explicit logging ref set is considered the default and should be unique as well)

**Expected behaviour**:
The logging operator should refuse to reconcile such a resource. It should update the logging resource status so the the user know what's wrong. Ideally the logging operator should have an admission webhook to reject such a second logging resource with the same ref.

**Steps to reproduce the bug**:
Anyone can create two logging resources without a logging ref set on either of those, but that should be avoided.

/kind bug

y-eight:
The same issue seems to occur if loggingRef is not set in different logging resources as well. FluentbitAgents dose not have a loggingRef set.

When reconciling the fluentbits from [loggingResources](https://github.com/kube-logging/logging-operator/blob/4.2.2/controllers/logging/logging_controller.go#L209) (FluentbitAgent List) all fluentbits that have an empty loggingRef in FluentbitAgent CR will be created, even if they are not part of the logging resource.

LoggingRed from FluentbitAgentList are compared [here](https://github.com/kube-logging/logging-operator/blob/4.2.2/pkg/resources/model/repository.go#L296)

Debugged with @eumel8

pepov:
I'm not sure I understand the problem statement here.

If there is no loggingRef defined on the resources, then all resources (Flow,Output,ClusterFlow,ClusterOutput,FluentbitAgent) will be considered as part of the logging resource that has no loggingRef defined either. Now if there are more than one logging resources defined with an empty loggingRef, than all Flow,Output,ClusterFlow,ClusterOutput,FluentbitAgent resources  will be considered part of all these loggings. This is expected and the root cause is still the same: the logging operator allows multiple logging resources with the same loggingRef.

Putting the fluentbit spec back into the logging resource does not solve this problem, it's still a totally invalid use case. Please *do not* use multiple logging resources without a uniq loggingRef!

Probably you are aware, but also please note there is a bug that has just been reported: https://github.com/kube-logging/logging-operator/issues/1388

pepov:
since this can be a breaking change, first we only introduce a deprecation notice

PR #1391, opened 2023-07-24
chore: deprecation notice for non-unique loggingRef

pepov:
Fixes https://github.com/kube-logging/logging-operator/issues/1386


controllers/logging/logging_controller.go
```diff
@@ -93,6 +93,28 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		if l.Name == logging.Name {
+			continue
+		}
+		if l.Spec.LoggingRef == logging.Spec.LoggingRef {
+			loggingsForTheSameRef = append(loggingsForTheSameRef, l.Name)
+		}
+	}
+
+	if len(loggingsForTheSameRef) > 0 {
+		problem := fmt.Sprintf("multiple other logging resources exist with the same loggingRef: %s",
```
siliconbrain, line 112:
```suggestion
		problem := fmt.Sprintf("other logging resources exist with the same loggingRef: %s",
```
might not be multiple others

controllers/logging/logging_controller.go
```diff
@@ -93,6 +93,28 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			strings.Join(loggingsForTheSameRef, ","))
+		logging.Status.Problems = []string{problem}
```
siliconbrain, line 114:
```suggestion
		logging.Status.Problems = append(logging.Status.Problems, problem)
```
this should be less error prone, but we should also clear the problem list since we're going to rebuild it as part of reconciliation

siliconbrain (approved):

controllers/logging/logging_controller.go
pepov: actually I wanted to update the logging resource right here since we return from the reconciliation right after this line, but I forgot...

pepov: completely restructure the code

pkg/resources/model/reconciler.go
```diff
@@ -203,6 +205,23 @@ func NewValidationReconciler(
...
+			if l.Name == resources.Logging.Name {
+				continue
+			}
+			if l.Spec.LoggingRef == resources.Logging.Spec.LoggingRef {
+				loggingsForTheSameRef = append(loggingsForTheSameRef, l.Name)
+			}
+		}
+
+		if len(loggingsForTheSameRef) > 0 {
+			problem := fmt.Sprintf("Deprecated behaviour! Other logging resources exist with the same loggingRef: %s",
```
aslafy-z, line 219: I think this message should announce what will happen on next major with these duplicated objects?

pepov: If only I knew! :) Just kidding, ok will add that

siliconbrain (approved):

PR #1392, opened 2023-07-24
watchNamespaces improvements

pepov:
- Make `logging.spec.watchNamespaces` (explicit static list) and `logging.spec.watchNamespaceSelector` (label based dynamic list) additive instead of exclusive for hopefully better semantics.
- Add tests for better coverage (looking into reusing the same thing for a different use case as well)
- Add namespace watches and trigger reconcile on loggings with watchNamespaceSelector defined

### DIFF
--- a/controllers/logging/logging_controller.go
+++ b/controllers/logging/logging_controller.go
@@ -107,10 +107,6 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		log.Info("WARNING PormetheusRule is not supported in the cluster")
 	}
 
-	if logging.Spec.WatchNamespaces != nil && logging.Spec.WatchNamespaceSelector != nil {
-		log.Info("WARNING watchNamespaceSelector will be omitted if configured along with watchNamespaces")
-	}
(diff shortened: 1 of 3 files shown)


pepov:
@nak0f as we are preparing to build on top of these selectors I wanted to reconsider making the two different namespace lists additive instead of exclusive to reduce frustration of users as they have to debug through the logging resource.

also this adds a few tests to cover the behaviour

aslafy-z:
An other option would be to remove completely `watchNamespaces` and encourage users to use the [`kubernetes.io/metadata.name` label (available since 1.22)](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#automatic-labelling) to select by name.

pepov:
@aslafy-z sounds good (I wasn't aware of that label), but that would be a breaking change affecting lots of users, thus I would avoid it now.

aslafy-z:
I opened https://github.com/kube-logging/logging-operator/issues/1394 to track this enhancement proposal.

siliconbrain (approved):
lgtm

PR #1396, opened 2023-07-26
helm: fix fluentd reference

aslafy-z:
helm templating fails with

```
Error: template: logging-operator/templates/logging.yaml:23:22: executing "logging-operator/templates/logging.yaml" at <.Values.logging.fluentd>: fluentd is not a method but has arguments Use --debug flag to render out invalid YAML
```

when `fluentd` value is set.

### DIFF
--- a/charts/logging-operator/templates/logging.yaml
+++ b/charts/logging-operator/templates/logging.yaml
@@ -20,29 +20,25 @@ spec:
   {{- end }}
   {{- if (not .Values.logging.fluentdDisabled) }}
   {{- if .Values.logging.fluentd }}
-  fluentd: {{- .Values.logging.fluentd . | nindent 4 }}
+  fluentd: {{- toYaml .Values.logging.fluentd | nindent 4 }}
   {{- else }}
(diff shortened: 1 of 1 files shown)


pepov (approved):

charts/logging-operator/templates/logging.yaml
```diff
@@ -20,7 +20,8 @@ spec:
   {{- end }}
   {{- if (not .Values.logging.fluentdDisabled) }}
   {{- if .Values.logging.fluentd }}
-  fluentd: {{- .Values.logging.fluentd . | nindent 4 }}
+  fluentd: 
+    {{- toYaml .Values.logging.fluentd | nindent 4 }}
```
pepov, line 24: why would you put it on a new line? nindent should do that for us

pepov: just a nit, lgtm anyways

aslafy-z: I copy pasted a line below. Just updated the whole file :)

PR #1399, opened 2023-07-26
add aggregated ClusterRole to manage namespaced resources

### LINKED ISSUE #1398: user roles for namespaced resources are missing in logging-operator helm chart
**Describe the bug**:

A user should have access to namespaced resources like Flows and Outputs. It seems this is not covered by the logging-operator Helm chart,only a global ClusterRole with ClusterRoleBinding will created. 

**Expected behaviour**:
User with admin permissions for a namespace should have access to Flows and Outputs of this namespace.

Try to get access to an existing Output:

```
 outputs.logging.banzaicloud.io "demoapp" is forbidden: User "u-xxxxxx" cannot get resource "outputs" in API group "logging.banzaicloud.io" in the namespace "demoapp"
```

/kind bug

eumel8:
fixes #1398

### DIFF
--- a/charts/logging-operator/templates/userrole.yaml
+++ b/charts/logging-operator/templates/userrole.yaml
@@ -14,6 +14,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
   - get
   - list
   - patch
@@ -35,6 +36,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
   - get
   - list
   - patch
@@ -57,6 +59,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
   - get
   - list
   - patch
--- a/charts/logging-operator/templates/userrole.yaml
+++ b/charts/logging-operator/templates/userrole.yaml
@@ -26,8 +26,6 @@ rules:
   - eventtailers/status
   verbs:
   - get
-  - patch
-  - update
 - apiGroups:
   - logging.banzaicloud.io
   resources:
@@ -49,8 +47,6 @@ rules:
   - outputs/status
   verbs:
   - get
-  - patch
-  - update
 - apiGroups:
   - logging.banzaicloud.io
   resources:
@@ -72,6 +68,4 @@ rules:
   - syslogngoutputs/status
   verbs:
   - get
-  - patch
-  - update
 {{- end }}
--- a/charts/logging-operator/templates/userrole.yaml
+++ b/charts/logging-operator/templates/userrole.yaml
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ template "logging-operator.fullname" . }}-user
+  name: {{ template "logging-operator.fullname" . }}-edit
   labels:
     rbac.authorization.k8s.io/aggregate-to-admin: "true"
 rules:
--- a/charts/logging-operator/templates/userrole.yaml
+++ b/charts/logging-operator/templates/userrole.yaml
@@ -25,13 +25,6 @@ rules:
 - apiGroups:
   - logging.banzaicloud.io
   resources:
-  - flows/status
-  - outputs/status
-  verbs:
-  - get
-- apiGroups:
-  - logging.banzaicloud.io
-  resources:
   - syslogngflows
   - syslogngoutputs
   verbs:
@@ -43,11 +36,4 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - logging.banzaicloud.io
-  resources:
-  - syslogngflows/status
-  - syslogngoutputs/status
-  verbs:
-  - get
 {{- end }}
--- a/charts/logging-operator/templates/userrole.yaml
+++ b/charts/logging-operator/templates/userrole.yaml
@@ -1,0 +1,74 @@
+{{- if .Values.rbac.enabled }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "logging-operator.fullname" . }}-user
+  labels:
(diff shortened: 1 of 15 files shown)


charts/logging-operator/templates/userrole.yaml
```diff
@@ -0,0 +1,74 @@
...
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups:
+  - logging-extensions.banzaicloud.io
+  resources:
+  - eventtailers
+  verbs:
+  - create
+  - delete
```
aslafy-z, line 16: Please add `deletecollection` to all resources

charts/logging-operator/templates/userrole.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - update
+  - watch
+- apiGroups:
+  - logging.banzaicloud.io
+  resources:
+  - syslogngflows/status
+  - syslogngoutputs/status
+  verbs:
+  - get
+  - patch
```
aslafy-z, line 72: Is `patch` and `update` verbs needed in here?

eumel8: good question. would that be an effect if a user updates the status manually?

pepov: the status is patched/updated by the operator exclusively, not something for the user to do.

pepov:
Thanks for the proposal and for the implementation! However I would like to separate these clusterroles based on the following resources:
- host tailer
- flows & outputs + syslogngflows & outputs
- clusterflows & outputs + syslogngclusterflows & outputs

I'm not sure about the eventtailer, since that is something to be installed once into a logging system, not something to manage by individual users.

I would add the edit/admin aggregation labels, so that it can easily be aggregated into a single clusterrole for editor/admin purposes, like it is done in the cert-manager chart:
https://github.com/cert-manager/cert-manager/blob/0b9366c0fb226393f2c51abdf9ec3097606a8713/deploy/charts/cert-manager/templates/rbac.yaml#L439

Also I would create a viewer role, that doesn't have to be split by resources, just grant view privileges to all the above logging resources for users.

I'm still a bit sceptical that we can provide clusterroles that are useful for most users/use cases, but at least we can provide an example implementation.

charts/logging-operator/templates/userrole.yaml
```diff
@@ -0,0 +1,77 @@
...
+kind: ClusterRole
+metadata:
+  name: {{ template "logging-operator.fullname" . }}-user
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups:
+  - logging-extensions.banzaicloud.io
+  resources:
+  - eventtailers
```
pepov, line 13: eventtailer is not a namespaced resource, I don't think we should add it here

charts/logging-operator/templates/userrole.yaml
eumel8: ok, removed

eumel8:
@pepov Rancher has this implementation also for [read-only/view access](https://github.com/rancher/charts/blob/release-v2.7/charts/rancher-logging/102.0.1%2Bup3.17.10/templates/userroles.yaml). 
Aggregation is a common RBAC feature in Kubernetes and useful not only in context with Rancher projects. An extra value flag under the rbac section of the chart is also possible. At the moment normal user haven't  access to Flow and Output resources.

charts/logging-operator/templates/userrole.yaml
```diff
@@ -0,0 +1,52 @@
+{{- if .Values.rbac.enabled }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "logging-operator.fullname" . }}-user
```
pepov, line 6:
```suggestion
  name: {{ template "logging-operator.fullname" . }}-edit
```

charts/logging-operator/templates/userrole.yaml
```diff
@@ -0,0 +1,52 @@
...
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```
pepov, line 8:
```suggestion
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```

charts/logging-operator/templates/userrole.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - logging.banzaicloud.io
+  resources:
+  - syslogngflows/status
+  - syslogngoutputs/status
+  verbs:
+  - get
```
aslafy-z, line 51: Is `watch` a verb that applies to status subresources?

charts/logging-operator/templates/userrole.yaml
aslafy-z: Looks like the edit ClusterRole is already aggregated to admin, I guess adding `aggregate-to-edit` here should be sufficient

charts/logging-operator/templates/userrole.yaml
pepov: I don't think so. Also I'm not entirely sure, why we need get on status. Do we need that at all?

eumel8: You mean the syslogng resource or all resources in general? Otherweise the user can't get access to the status.

pepov: I mean if you have get access on a resource, you should have get access on the subresource as well.

eumel8: Yes, make sense. Tested and the status is editable in any way, but the Operator sets back.

pepov (approved):

PR #1407, opened 2023-07-31
make a full fluentd run for configcheck

eumel8:
We have still the problem that the fluentd configcheck 1.15 didn't correct handled the check for elasticsearch version.
Elasticsearch version 7 targets passed the configcheck and fluentd crashs due the connection error between elasticsearch client (version 8) and the Output target (version 7). A full run of fluentd would end in error state of the configcheck and the fluentd would still running. Other checks  may also be affected.


pkg/resources/fluentd/appconfigmap.go
```diff
@@ -312,9 +312,9 @@ func (r *Reconciler) volumesCheckPod(hashKey string) (v []corev1.Volume) {
 
 func (r *Reconciler) containerCheckPod(hashKey string) []corev1.Container {
 	containerArgs := []string{
+		"timeout", "10",
 		"fluentd", "-c",
 		fmt.Sprintf("/fluentd/etc/%s", ConfigKey),
-		"--dry-run",
```
eumel8, line 317: With considerations, to make this args configurable, I came to this conclusion, that `extraArgs` already exists. So, a user can add `--dry-run` on this place. Only the timeout command is required to exit fluentd in any case.

pepov: Well, it is, however I think it would worth it's own configuration structure so that it's straightForward to configure through the API.

Something like:
```
spec:
  fluentd:
    configCheckStrategy: DryRun|Timeout
    configCheckTimeoutSeconds: 10 
```

I'm also thinking about whether this makes sense for syslog-ng or not, because then we can make it a common aggregator configuration as well.

pepov: Yes, we definitely want this for syslog-ng as well. Also I'm considering creating a feature request in syslog-ng to provide a flag to exit after initialization, we could do something similar for fluentd as well.

pepov:
Thanks for your contribution @eumel8 ! The idea has been used and extended in https://github.com/kube-logging/logging-operator/pull/1431

PR #1409, opened 2023-08-02
[docs] Order types alphabetically

eumel8:
CRD docs are hard to read if values are not orderly. The user need to read the full page to find out configuration options.  
Let's order alphabetically, so same options are together and the full page has a better readable.


pepov:
In my opinion docs generation should handle the alphabetical ordering and should not be enforced through go code and manual reviews.

eumel8:
In general I would say: Yes, a kind of automation should build things in the right order. It seems there is [already](https://github.com/cisco-open/operator-tools/blob/master/pkg/docgen/lister.go#L138) some implemented. On the other hand API specs are growing with the time and it's a bit confusing if they are not in the alphabetically order, which is not so good to read and understand. 
Not sure how to proceed here before solve the merge conflicts.

pepov:
Closing this for now but added https://github.com/kube-logging/logging-operator/issues/1468 to track the request

PR #1410, opened 2023-08-02
introduce BufferVolumeResources

eumel8:
The bufferMetricsSidecarContainer has no resource configuration option at the moment. Extend this with the BufferVolumeResources

### DIFF
--- a/pkg/resources/fluentd/statefulset.go
+++ b/pkg/resources/fluentd/statefulset.go
@@ -84,7 +84,7 @@ func (r *Reconciler) statefulsetSpec() *appsv1.StatefulSetSpec {
 		fluentContainer(r.Logging.Spec.FluentdSpec),
 		*newConfigMapReloader(r.Logging.Spec.FluentdSpec),
 	}
-	if c := r.bufferMetricsSidecarContainer(); c != nil {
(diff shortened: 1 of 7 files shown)


pkg/resources/fluentd/statefulset.go
```diff
@@ -364,7 +364,7 @@ func (r *Reconciler) volumeMountHackContainer() *corev1.Container {
 	return nil
 }
 
-func (r *Reconciler) bufferMetricsSidecarContainer() *corev1.Container {
+func (r *Reconciler) bufferMetricsSidecarContainer(spec *v1beta1.FluentdSpec) *corev1.Container {
```
pepov, line 367: the spec is already avalable under `r.Logging.Spec.FluentdSpec`

pepov:
Another pr was also submitted for this here:
https://github.com/kube-logging/logging-operator/pull/1426

I summarize my comments here as well.

- We need this configuration for fluentd and fluentbit too, which has the same buffer sidecar
- We need sane defaults so that it should work out of the box.

We need defaults for syslog-ng as well, but I'm happy to contribute that one once we have the above.

pepov:
work continued on https://github.com/kube-logging/logging-operator/pull/1426

thank you @eumel8 for taking the time for this!

PR #1414, opened 2023-08-02
feat: fluentbit hotreload

### LINKED ISSUE #1312: Fluentbit config hot reload
Fluentbit restarts can take a long time if there is lot of files to index. 

Integration of the fluentbit new hot reload feature, see
https://fluentbit.io/announcements/v2.1.0/

aslafy-z:
Enable [fluentbit hotreload feature](https://docs.fluentbit.io/manual/administration/hot-reload) behind a feature flag (FluentbitSpec.HotReload).

I'd say it misses an E2E test. Please do an initial review first. :pray:  

Closes #1312


pepov:
Looks good, but I would be conservative to keep at least the option of switching back to the legacy method.

pkg/resources/fluentbit/daemonset.go
```diff
@@ -147,6 +154,44 @@ func (r *Reconciler) generatePortsMetrics() (containerPorts []corev1.ContainerPo
 	return
 }
 
+func newConfigMapReloader(spec *v1beta1.FluentbitSpec) *corev1.Container {
```
pepov, line 157: can't we simply return by value?

pkg/sdk/logging/api/v1beta1/fluentbit_types.go
```diff
@@ -80,13 +81,14 @@ type FluentbitSpec struct {
...
-	CoroStackSize int32                       `json:"coroStackSize,omitempty" plugin:"default:24576"`
-	Resources     corev1.ResourceRequirements `json:"resources,omitempty"`
-	Tolerations   []corev1.Toleration         `json:"tolerations,omitempty"`
-	NodeSelector  map[string]string           `json:"nodeSelector,omitempty"`
-	Affinity      *corev1.Affinity            `json:"affinity,omitempty"`
-	Metrics       *Metrics                    `json:"metrics,omitempty"`
-	Security      *Security                   `json:"security,omitempty"`
+	CoroStackSize           int32                       `json:"coroStackSize,omitempty" plugin:"default:24576"`
+	Resources               corev1.ResourceRequirements `json:"resources,omitempty"`
+	ConfigReloaderResources corev1.ResourceRequirements `json:"configReloaderResources,omitempty"`
```
pepov, line 86: I would consider a different structure here:
```
configHotReload: # pointer, which means hotreload is enabled if set
  image:
  resources:
```

aslafy-z: I replicated the pattern used for aggregators. Do you want to migrate them to this spec too?

pepov: I would not touch that, but I also don't want to add 3 new fields to the already big fluentbit spec. I understand that consistency is a nice thing, but I would rely on the user having to read the docs in this case.

pkg/resources/fluentbit/daemonset.go
aslafy-z: Done

pkg/sdk/logging/api/v1beta1/fluentbit_types.go
aslafy-z: How can I prevent kubebuilder from adding these subfields into the CRD?

pepov: which one specifically?

aslafy-z: Content of the HotReload struct

pkg/resources/fluentbit/daemonset.go
```diff
@@ -147,6 +154,44 @@ func (r *Reconciler) generatePortsMetrics() (containerPorts []corev1.ContainerPo
...
+	var args []string
+	vm := []corev1.VolumeMount{
+		{
+			Name:      "config",
+			MountPath: OperatorConfigPath,
+		},
+	}
+
+	args = append(args,
+		"--volume-dir=/fluentbit/config",
```
pepov, line 167: shouldn't this be `OperatorConfigPath`?

pkg/sdk/logging/api/v1beta1/fluentbit_types.go
pepov: I don't think you can reference an object in a crd spec, do you have an example where you have seen that working?

aslafy-z: > I also don't want to add 3 new fields to the already big fluentbit spec
Any idea how I can skip these fields in the crd?

pepov: I meant I would avoid adding 3 fields on the top level. This is all good as it is now.

pkg/resources/fluentbit/daemonset.go
pepov: @aslafy-z are you sure this is the right path?

aslafy-z, line 167:
```suggestion
		fmt.Sprintf("--volume-dir=%s", OperatorConfigPath),
```

aslafy-z: Nice catch. It's still untested!

pepov:
Since LoggingRoute is now in, we could test this by adding a LoggingRoute for a specific Logging resource as we go and check it being applied live.

aslafy-z:
@pepov I added an E2E test. Can you please give a look? :)

pkg/resources/fluentbit/daemonset.go
```diff
@@ -58,6 +58,9 @@ func (r *Reconciler) daemonSet() (runtime.Object, reconciler.DesiredState, error
 	containers := []corev1.Container{
 		*r.fluentbitContainer(),
 	}
+	if r.fluentbitSpec.ConfigHotReload != nil {
+		containers = append(containers, newConfigMapReloader(r.Logging.Spec.FluentbitSpec))
```
aslafy-z, line 62:
```suggestion
		containers = append(containers, newConfigMapReloader(r.fluentbitSpec))
```

e2e/fluentbit-hotreload/fluentbit_hotreload_test.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+	}
+	common.RequireNoError(t, c.Create(ctx, &flow))
+
+	agent := v1beta1.FluentbitAgent{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "infra",
+		},
+		Spec: v1beta1.FluentbitSpec{
+			LoggingRef: "infra",
```
aslafy-z, line 212:
```suggestion
			LoggingRef: "infra",
			ConfigHotReload: map[string]string{},
```

e2e/fluentbit-hotreload/fluentbit_hotreload_test.go
```diff
@@ -0,0 +1,323 @@
+// Copyright © 2021 Cisco Systems, Inc. and/or its affiliates
```
pepov, line 1:
```suggestion
// Copyright © 2024 Kube logging authors, Inc. and/or its affiliates
```

pepov (commented):
Hugo kudos for the test! I would like to add a few things there if you don't mind:
- check that the fluentbit daemonset has been start
- check that no logs are coming before adding the loggingroute
- check if the daemonset has the same generation after adding the loggingroute

I'm happy to put these on top of your branch, just let me know if you have any concerns.

pepov (approved):

OverOrion (approved):
LGTM, thanks for this! :tada: :raised_hands:

PR #1415, opened 2023-08-02
Fix docs generation

fekete-robert:
Because it breaks generating the reference doc (make docs)


pepov:
That is not part of the API and no docs should be generated actually. We should skip it or move to another file perhaps?

fekete-robert:
I'm not sure how to skip it, I'll try to check operator-tools next week

pepov:
fixed by moving the struct in question to a more relevant package and restored docs check in CI

pepov (approved):

aslafy-z (approved):

PR #1417, opened 2023-08-02
fix(fluentd): use aggregated metrics with multi workers

### LINKED ISSUE #1110: Multiworker support for aggregated_metrics endpoint in servicemonitor

aslafy-z:
Select `/aggregated_metrics` Fluentd metrics endpoint by default when multi worker is enabled.
Add E2E test that checks if multi worker is enabled and the two workers correctly get logs.

Tested:
- explicitly setting metrics.path to `/metrics` changes servicemonitor and prom annotation to point to `/metrics` independently of the worker count
- if metrics.path is not set and workers = 1, servicemonitor and prom annotations point to `/metrics`
- if metrics.path is not set and workers > 1, servicemonitor and prom annotations point to `/aggregated_metrics`

Fixes #1110

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ test: generate fmt vet manifests ${ENVTEST_BINARY_ASSETS} ${KUBEBUILDER} ## Run 
 
 .PHONY: test-e2e
 test-e2e: ${KIND} generate manifests docker-build stern ## Run E2E tests
-	$(MAKE) test-e2e-nodeps
+	$(MAKE) test-e2e-nodeps E2E_TEST=${E2E_TEST}
 
 .PHONY: test-e2e-ci
 test-e2e-ci: ${BIN}
@@ -164,12 +164,16 @@ test-e2e-ci: ${BIN}
 	chmod +x ./bin/kind
 	curl -L https://github.com/stern/stern/releases/download/v${STERN_VERSION}/stern_${STERN_VERSION}_linux_amd64.tar.gz | tar xz -C bin stern
 	chmod +x ./bin/stern
+	$(MAKE) test-e2e-nodeps E2E_TEST=${E2E_TEST}
+
+.PHONY: test-e2e-nodeps
+test-e2e-nodeps:
 	cd e2e && \
 		LOGGING_OPERATOR_IMAGE="${IMG}" \
 		KIND_PATH="$(KIND)" \
 		KIND_IMAGE="$(KIND_IMAGE)" \
 		PROJECT_DIR="$(PWD)" \
-		go test -v -timeout ${E2E_TEST_TIMEOUT} ./...
+		go test -v -timeout ${E2E_TEST_TIMEOUT} ./${E2E_TEST}/...
 
 .PHONY: tidy
 tidy: ## Tidy Go modules
--- a/e2e/fluentd-aggregator/fluentd_aggregator_test.go
+++ b/e2e/fluentd-aggregator/fluentd_aggregator_test.go
@@ -1,0 +1,216 @@
+// Copyright © 2021 Cisco Systems, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fluentd_aggregator
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/cisco-open/operator-tools/pkg/utils"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
+
+	"github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
+	"github.com/kube-logging/logging-operator/pkg/sdk/logging/model/output"
+
+	"github.com/kube-logging/logging-operator/e2e/common"
+	"github.com/kube-logging/logging-operator/e2e/common/cond"
+	"github.com/kube-logging/logging-operator/e2e/common/setup"
+)
+
+var TestTempDir string
+
+func init() {
+	var ok bool
+	TestTempDir, ok = os.LookupEnv("PROJECT_DIR")
+	if !ok {
+		TestTempDir = "../.."
+	}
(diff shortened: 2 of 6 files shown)


pkg/resources/fluentd/config.go
```diff
@@ -54,7 +54,6 @@ var fluentdInputTemplate = `
 <source>
     @type prometheus
     port {{ .Monitor.Port }}
-    metrics_path {{ .Monitor.Path }}
```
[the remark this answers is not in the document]
pepov: But how do we notify users who accidentally configured this to something different than `/metrics` in the default case?

I see a few options here:
- configure the annotation and the fluentd config differently (leave the fluentd config empty) and update the template to configure the path only if it's set.
- update the template to avoid setting the path if it's explicitly `/metrics or /aggregated_metrics`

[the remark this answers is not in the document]
aslafy-z: I did the first option. Let me know what you think about it.

Makefile
```diff
@@ -76,7 +76,7 @@ deploy: manifests ## Deploy controller in the configured Kubernetes cluster in ~
 
 .PHONY: docker-build
 docker-build: ## Build the docker image
-	docker image inspect ${IMG} && echo "docker image ${IMG} already built" || ${DOCKER} build . -t ${IMG}
```
aslafy-z, line 79: This was skipping the build on code change. Docker already caches the layers. Ok for you to remove this check?

Makefile
```diff
@@ -156,20 +156,24 @@ test: generate fmt vet manifests ${ENVTEST_BINARY_ASSETS} ${KUBEBUILDER} ## Run
...
 .PHONY: test-e2e-ci
 test-e2e-ci: ${BIN}
 	curl -Lo ./bin/kind https://kind.sigs.k8s.io/dl/${KIND_VERSION}/kind-linux-amd64
 	chmod +x ./bin/kind
 	curl -L https://github.com/stern/stern/releases/download/v${STERN_VERSION}/stern_${STERN_VERSION}_linux_amd64.tar.gz | tar xz -C bin stern
 	chmod +x ./bin/stern
+	$(MAKE) test-e2e-nodeps E2E_TEST=${E2E_TEST}
+
+.PHONY: test-e2e-nodeps
+test-e2e-nodeps:
```
aslafy-z, line 170: This step was missing. I think this is what was planned.

Makefile
```diff
@@ -156,20 +156,24 @@ test: generate fmt vet manifests ${ENVTEST_BINARY_ASSETS} ${KUBEBUILDER} ## Run
...
 	cd e2e && \
 		LOGGING_OPERATOR_IMAGE="${IMG}" \
 		KIND_PATH="$(KIND)" \
 		KIND_IMAGE="$(KIND_IMAGE)" \
 		PROJECT_DIR="$(PWD)" \
-		go test -v -timeout ${E2E_TEST_TIMEOUT} ./...
+		go test -v -timeout ${E2E_TEST_TIMEOUT} ./${E2E_TEST}/...
```
aslafy-z, line 176: This variable allows launching a test suite specifically using makefile arguments:
`make test-e2e E2E_TEST=fluent-aggregator`

pkg/resources/fluentd/configsecret.go
```diff
@@ -56,8 +56,13 @@ func (r *Reconciler) generateConfigSecret() (map[string][]byte, error) {
 	input := fluentdConfig{
 		IgnoreSameLogInterval:     r.Logging.Spec.FluentdSpec.IgnoreSameLogInterval,
 		IgnoreRepeatedLogInterval: r.Logging.Spec.FluentdSpec.IgnoreRepeatedLogInterval,
-		RootDir:                   r.Logging.Spec.FluentdSpec.RootDir,
 		EnableMsgpackTimeSupport:  r.Logging.Spec.FluentdSpec.EnableMsgpackTimeSupport,
+		Workers:                   r.Logging.Spec.FluentdSpec.Workers,
+	}
+
+	input.RootDir = r.Logging.Spec.FluentdSpec.RootDir
```
aslafy-z, line 63: This default value is computed at this level so I can use the already defined `bufferPath` variable. 
Let me know if you want me to move it into the `logging_types.go` file.

e2e/fluentd-aggregator/fluentd_aggregator_test.go
```diff
@@ -0,0 +1,276 @@
...
+			rawOut, err := cmd.Output()
+			if err != nil {
+				t.Logf("failed to get log consumer logs: %v", err)
+				return false
+			}
+			t.Logf("log consumer logs: %s", rawOut)
+			return strings.Contains(string(rawOut), testTag)
+		}, 5*time.Minute, 3*time.Second)
+
+		cmd := common.CmdEnv(exec.Command("kubectl", "scale",
```
pepov, line 197: test code after this is unnecessary for this test case I think

pkg/resources/fluentd/configsecret.go
pepov: It would make sense, thanks for pointing this out!
Could you please also move the loglevel default while you are here?

aslafy-z: Done with [`25f0a78` (#1417)](https://github.com/kube-logging/logging-operator/pull/1417/commits/25f0a7880062e17da06be64195d2dd86165f031c)

e2e/fluentd-aggregator/fluentd_aggregator_test.go
aslafy-z: fixed in [`a1db6c0` (#1417)](https://github.com/kube-logging/logging-operator/pull/1417/commits/a1db6c01290062a57acf0634684343195272f8fd)

Makefile
aslafy-z: Reverted

pepov (approved):

pepov:
I don't want to hold this PR any further, so created a follow up for proper e2e tests on the metrics endpoints too, which we don't have currently: https://github.com/kube-logging/logging-operator/issues/1427

PR #1431, opened 2023-08-14
configcheck fullrun timeout

pepov:
- make a full fluentd run for configcheck
- chore(make): remove obsolete make target
- feat: add timeout based configcheck for both forwarders


aslafy-z (approved):

OverOrion (approved):
LGTM!

PR #1438, opened 2023-08-18
feat: extra options to filetailer

aslafy-z:


aslafy-z:
These options are not supported by the CLI. I opened an issue at https://github.com/kube-logging/logging-operator/issues/1440 to move to a configmap approach.

PR #1439, opened 2023-08-18
docs: add sample for syslog-ng

aslafy-z:


pepov (approved):

config/samples/syslog-ng.yaml
```diff
@@ -0,0 +1,27 @@
+kind: Logging
+apiVersion: logging.banzaicloud.io/v1beta1
+metadata:
+  name: logging
+spec:
+  syslogNG:
```
aslafy-z, line 6: @pepov NG in upper case ?

pepov: yes

config/samples/syslog-ng.yaml
```diff
@@ -0,0 +1,27 @@
...
+    bufferVolumeMetrics:
+      mount_name: buffers
+      serviceMonitor: true
```
pepov, line 9:
```suggestion
      # prometheus chart or the operator should be installed for this to work
      serviceMonitor: true
```

PR #1441, opened 2023-08-20
feat: add elasticsearch use_legacy_template option

### LINKED ISSUE #1034: Support new Elasticsearch "composable" index template format

I am using the logging operator (helm chart) version 3.17.7 with the Elasticsearch output plugin for Fluentd. It would seem that it uses the deprecated Elastic [legacy index template](https://www.elastic.co/guide/en/elasticsearch/reference/7.13/indices-templates-v1.html) rather than the newer [composable index templates](https://www.elastic.co/guide/en/elasticsearch/reference/7.13/index-templates.html).

The problem is that I have is that it does not appear to be possible to configure the logging operator to use the newer composable index templates.

I've written this as a feature request because I am able to use the logging operator with legacy templates fine, so this isn't a bug. However **legacy templates are deprecated** so a way of using the composable templates is desirable.

Although the plugin supports a [use_legacy_template](https://github.com/uken/fluent-plugin-elasticsearch#use_legacy_template) switch this does not appear to be recognised by the CRD `io.banzaicloud.logging.v1beta1.ClusterOutput.spec.elasticsearch` (it's documented in the plugin but not on the [logging operator plugin page](https://banzaicloud.com/docs/one-eye/logging-operator/configuration/plugins/outputs/elasticsearch)).

There is a warning output to the fluentd log about the deprecation
```
fluentd warning: 299 Elasticsearch-7.13.2-4d960a0733be83dd2543ca018aa4ddc42e956800 "Legacy index templates are deprecated in favor of composable templates."
```

I would like to be able to add `use_legacy_template: false` to the Elasticsearch output spec so that the plugin works with composable index templates.

I did try to add `use_legacy_template: false` to a `clusterOutput` spec anyway:

```
# clusteroutputs.logging.banzaicloud.io "elasticsearch" was not valid:
# * <nil>: Invalid value: "The edited file failed validation": ValidationError(ClusterOutput.spec.elasticsearch): unknown field "use_legacy_template" in io.banzaicloud.logging.v1beta1.ClusterOutput.spec.elasticsearch
```

I also manually edited the generated `fluentd.conf` to include this feature and this did change the plugin to use the composable templates. However I didn't follow through with a working configuration because I need to drive it from the logging operator CRD - therefore I can't comment on whether the plugin actually works with composable templates (although I expect it likely does, given that it's documented).

aslafy-z:
Fixes #1034


pkg/sdk/logging/model/output/elasticsearch.go
```diff
@@ -259,6 +259,9 @@ type ElasticsearchOutput struct {
 	DataStreamIlmPolicy string `json:"data_stream_ilm_policy,omitempty"`
 	// Specify whether overwriting data stream ilm policy or not.
 	DataStreamIlmPolicyOverwrite bool `json:"data_stream_ilm_policy_overwrite,omitempty"`
+	// Specify wether to use legacy template or not. (default: true)
+	// +kubebuilder:validation:Optional
+	UseLegacyTemplate *bool `json:"use_legacy_template,omitempty" plugin:"default:true"`
```
pepov, line 264: I beleive we don't have to add ```plugin:"default:true"``` if this is the default in the output anyways. It will make an unnecessary configcheck and reload for existing deployments.

pepov: Checked once again, this seems to be the default in the fluentd plugin, so we should omit the plugin tag on our side to avoid unecessary config changes in existing deployments: https://github.com/uken/fluent-plugin-elasticsearch/blame/28ade35649cff7a2dd49dd550ac6576cffde8af0/lib/fluent/plugin/out_elasticsearch.rb#L176

pepov: @aslafy-z in case you missed this one

aslafy-z: Thank you, I missed it. Applied!

pepov (approved):

pkg/sdk/logging/model/output/elasticsearch_test.go
```diff
@@ -96,6 +97,7 @@ buffer:
     scheme https
     ssl_verify false
     ssl_version TLSv1_2
+    use_legacy_template true
```
pepov, line 100:
```suggestion
```

pkg/sdk/logging/model/output/elasticsearch_test.go
```diff
@@ -48,6 +48,7 @@ buffer:
 	scheme https
 	ssl_verify false
 	ssl_version TLSv1_2
+	use_legacy_template true
```
pepov, line 51:
```suggestion
```

pepov (changes requested):

pepov (approved):

pepov:
I just fixed the tests, for me this is good to go

OverOrion (approved):
LGTM, thanks for this!

Issue #1368, opened 2023-06-21
Renaming the default Fluentd buffer PVC


cbuto:
**Is your feature request related to a problem? Please describe.**
When deploying the logging-operator in GCP using the GCP CSI driver, its possible to run into limitations due to the default name of the Fluentd buffer PVC.

Example error from the CSI driver:
```
...
failed to insert zonal disk: unknown Insert disk error: googleapi: Error 400: Invalid value for field 'resource.labels': ''. Label value 'logging-operator-logging-fluentd-buffer-logging-operator-logging-fluentd-0' violates format constraints. The value can only contain lowercase letters, numeric characters, underscores and dashes. The value can be at most 63 characters long. International characters are allowed., invalid
...
```

Unfortunately, its not possible to override this name like:
```
bufferStorageVolume:
  pvc:
    source:
      # update the name here
      claimName: buf
    spec:
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
      volumeMode: Filesystem
```

This is because updating the `claimName` results in an error in the statefulset when the `buffer-metrics-sidecar` is enabled:
```
 "logging-operator-logging-fluentd-0" is invalid: spec.containers[2].volumeMounts[0].name: Not found: "logging-operator-logging-fluentd-buffer"
```

I believe this is because that volumeMount name is hardcoded to the default of `fluentd-buffer`:
https://github.com/kube-logging/logging-operator/blob/e183d85815b3111da679704101a3c5b7f6d28b02/pkg/resources/fluentd/statefulset.go#L389-L390

**Describe the solution you'd like**
When the `fluentd.bufferStorageVolume.pvc.source.claimName` name is set, the buffer metrics sidecar container should be updated to use the new volume mount name.

**Describe alternatives you've considered**
N/A

**Additional context**
Somewhat related to https://github.com/kube-logging/logging-operator/issues/1327

cbuto:
@pepov opened up https://github.com/kube-logging/logging-operator/pull/1443 with a small fix to use the proper volume mount name for the buffer metrics sidecar. let me know what you think

PR #1443, opened 2023-08-23
fix(fluentd): Use buffer volume mount name from spec instead of default

cbuto:
Fixes https://github.com/kube-logging/logging-operator/issues/1368

This PR updates the `fluentd` buffer metrics sidecar to use the configured buffer volume name instead of the default. Prior to this change, updating the `bufferStorageVolume.pvc.source.claimName` would result in the buffer metrics sidecar attempting to mount in the wrong volume name because of the hardcoded const value.

tested with:
```
kind: Logging
metadata:
  name: all-to-file
spec:
  fluentd:
    bufferStorageVolume:
      pvc:
        source:
          claimName: buf
        spec:
          accessModes:
            - ReadWriteOnce
          resources:
            requests:
              storage: 10Gi
          volumeMode: Filesystem
    bufferVolumeMetrics:
      port: 9200
      path: /metrics
      prometheusAnnotations: true
  controlNamespace: default
  fluentbit: {}
```


pepov:
@cbuto Nice, but please add a check to see if we are dealing with a pvc or not (not flexible but should still be supported): https://github.com/kube-logging/logging-operator/commit/54fdce9672cde3fb0568dd590ed16d621c85c140

cbuto:
> @cbuto Nice, but please add a check to see if we are dealing with a pvc or not (not flexible but should still be supported): [54fdce9](https://github.com/kube-logging/logging-operator/commit/54fdce9672cde3fb0568dd590ed16d621c85c140)

@pepov thanks, good call. I didn't know that was an option 😄

OverOrion (approved):
LGTM, thank you for contributing!

PR #1445, opened 2023-08-25
feat(syslogng-http-output): add syslog-ng config fields to logging-operator output spec

kristofgyuracz:
- Add ` log-fifo-size`, `response-action` and `timeout` field for HTTP Output
- Fix behaviour of `peer_verify` field


pkg/sdk/logging/model/syslogng/output/http.go
```diff
@@ -65,8 +66,11 @@ type HTTPOutput struct {
 	// The value of the USER-AGENT header in the messages sent to the server.
 	UserAgent string `json:"user-agent,omitempty"`
 	// Description: Specifies the number of worker threads (at least 1) that syslog-ng OSE uses to send messages to the server. Increasing the number of worker threads can drastically improve the performance of the destination.
-	Workers     int    `json:"workers,omitempty"`
-	PersistName string `json:"persist_name,omitempty"`
+	Workers        int                `json:"workers,omitempty"`
+	PersistName    string             `json:"persist_name,omitempty"`
+	LogFifoSize    int                `json:"log-fifo-size,omitempty"`
```
siliconbrain, line 71:
```suggestion
	LogFIFOSize    int                `json:"log-fifo-size,omitempty"`
```
follow naming guidelines https://github.com/golang/go/wiki/CodeReviewComments#initialisms

siliconbrain (approved):

pkg/sdk/logging/model/syslogng/output/tls.go
```diff
@@ -41,7 +41,7 @@ type TLS struct {
 	// Name of a file, that contains an X.509 certificate (or a certificate chain) in PEM format, suitable as a TLS certificate, matching the private key set in the key-file() option. [more information](https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.37/administration-guide/73#kanchor3146)
 	CertFile *secret.Secret `json:"cert_file,omitempty"`
 	// Verification method of the peer. [more information](https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.37/administration-guide/73#tls-options-peer-verify)
```
orymate, line 43:
```suggestion
	// Verification method of the peer. [more information](https://axoflow.com/docs/axosyslog-core/chapter-encrypted-transport-tls/tlsoptions/#tls-options-peer-verify)
```

404

orymate:
pls regenerate the docs (for the changed crd)

pkg/sdk/logging/model/syslogng/output/tls.go
```diff
@@ -41,7 +41,7 @@ type TLS struct {
...
-	PeerVerify string `json:"peer_verify,omitempty"`
+	PeerVerify *bool `json:"peer_verify,omitempty"`
```
pepov, line 44: wasn't `peer_verify` (with underscore) invalid?

kristofgyuracz: actually it's valid, i was just misinterpreting the error message.

kristofgyuracz: but if we want, we can still change this field to use a dash, but i didn't want to introduce unnecessary (breaking) api changes.

OverOrion: I would say that we should leave it as is, because both the underscore and dash versions can be found in the codebase now, and once there is a an agreed preference then we should do big formatting PR.

OverOrion (approved):
LGTM, thank you for this!

PR #1449, opened 2023-08-31
feat(api): Provide way to disable istio sidecar injection with labels

kristofgyuracz:
The changes ensure that no istio sidecar is injected by default into jobs/individual pods that run to completion

See: https://github.com/istio/istio/issues/6324

- Config checkers, and fluentd drainer pods now have `sidecar.istio.io/inject` set to `false` by default.
- Fluentd drainer labels can be configured in the `Logging` spec.


pkg/sdk/logging/api/v1beta1/fluentd_types.go
```diff
@@ -148,8 +148,10 @@ type FluentdTLS struct {
 type FluentdDrainConfig struct {
 	// Should buffers on persistent volumes left after scaling down the statefulset be drained
 	Enabled bool `json:"enabled,omitempty"`
-	// Container image to use for the drain watch sidecar
+	// Annotations to use for the drain watch sidecar
 	Annotations map[string]string `json:"annotations,omitempty"`
+	// Labels to use for the drain watch sidecar
```
pepov, line 153:
```suggestion
	// Labels to use for the drain watch sidecar on top of labels added by the operator by default
```

pepov (approved):

pkg/sdk/logging/api/v1beta1/fluentd_types.go
kristofgyuracz: Currently, users can also override the default labels. If that is the right behaviour, we should also add that to the comment, otherwise i should change the logic.

pepov: yes, my intention was that, feel free to change the comment or ignore, it's just a nit

aslafy-z:
Everyone is not an istio user. I guess there should not be such a default value.

kristofgyuracz:
For non-istio users, these default should make no difference, this label is only used in istio context.
For istio users, these defaults make logging-operator work out of the box, without the need for an istio-specific documentation article, and manuals steps.
In the long term, k8s already has improvements in progress to handle sidecars differently from "workload" containers. With that, istio should be able to handle this issue without disabling sidecar injection.

pkg/resources/syslogng/configcheck.go
```diff
@@ -266,3 +267,12 @@ func (r *Reconciler) newCheckPod(hashKey string) (*corev1.Pod, error) {
 func configCheckResourceName(hash string) string {
 	return fmt.Sprintf("syslog-ng-configcheck-%s", hash)
 }
+
+func (r *Reconciler) configCheckPodObjectMeta(name, component string) metav1.ObjectMeta {
+	objectMeta := r.SyslogNGObjectMeta(name, component)
+
+	// Ensure that no istio sidecar is injected by default. See: https://github.com/istio/istio/issues/6324
```
OverOrion, line 274: Could you please add this to the PR description as well? Just so if someone clicks on the PR they will see it immediately without going through the changes.

OverOrion (approved):
LGTM, thanks for this!

pkg/resources/syslogng/configcheck.go
pepov: adding it now

pepov:
> Everyone is not an istio user. I think there should not be such default values.

@aslafy-z you are right, although I think we can and should support tools with dominant adoption for certain use cases by default, like we do support prometheus crds for monitoring (although not necessarily everyone is a prometheus user).

So I'm not for or against this, but one thing I haven't noticed is that for the configcheck there is no option to override or to add labels (for other service mesh implementations) so that should be added

pepov (changes requested):
Please add label override option for the configcheck as well. We have a configCheck field in the logging spec, it can be used for both aggregators

pepov (approved):

aslafy-z (approved):
LGTM 🙏

PR #1451, opened 2023-08-31
Update syslog-ng configs and metrics exporter image

pepov:
- Add configuration to set log level
- Update stats defaults and stricter behaviour over deprecated fields
- Static and dynamic defaults for max-connections
- Update metrics exporter image


orymate (approved):

pkg/sdk/logging/api/v1beta1/syslogng_types.go
```diff
@@ -66,7 +66,10 @@ type GlobalOptions struct {
 	// deprecated use stats/freq from 4.1+
 	StatsFreq *int `json:"stats_freq,omitempty"`
 	// TODO switch to this by default
```
OverOrion, line 68: Is this TODO item still valid?

pepov: no and it's not a good idea to write todos into places that we use to generate docs from (note for myself)
fixed in 67a352c3

kristofgyuracz (approved):
LGTM.

PR #1453, opened 2023-09-01
Check docs folder for diff in CI

OverOrion:
The `./docs/` folder was not checked previously, this PR aims to fix it.

The first commit will fail, and the second commit should add the missing docs files.


pepov (approved):

OverOrion:
I am unsure why the CI is green at the moment as there are docs that have not been generated yet (elasticsearch).

OverOrion:
I am going to break the CI in the next commit, and then revert it, just to make sure it actually works.

kristofgyuracz (approved):
LGTM.

PR #1455, opened 2023-09-04
Add mongodb output to syslog-ng

OverOrion:
Continuation of https://github.com/kube-logging/logging-operator/pull/1255.

My question is about the `write-concern` option, which can take a builtin keyword or a non-negative integer (the [docs](https://axoflow.com/docs/axosyslog-core/chapter-destinations/configuring-destinations-mongodb/reference-destination-mongodb/#mongodb-option-write-concern) does not say the latter for some reason, might need to be looked into). In the last commit I refactored it to use an interface, however the ergonomics are still not great, is there any way to somehow improve it, just leave as is or even revert back to the `RawString` version?


pepov:
I don't see too much value added through the WriteConcern interface to be honest, so I would opt for the basic rawstring solution.

pkg/sdk/logging/model/syslogng/output/mongodb.go
```diff
@@ -0,0 +1,116 @@
...
+	LogFIFOSize int `json:"log-fifo-size,omitempty"`
+	// If you receive the following error message during AxoSyslog startup, set the persist-name() option of the duplicate drivers:
+	// `Error checking the uniqueness of the persist names, please override it with persist-name option. Shutting down.`
+	// See [syslog-ng docs](https://axoflow.com/docs/axosyslog-core/chapter-destinations/configuring-destinations-http-nonjava/reference-destination-http-nonjava/#persist-name) for more information.
+	PersistName string `json:"persist_name,omitempty"`
+	// The number of times syslog-ng OSE attempts to send a message to this destination. If syslog-ng OSE could not send a message, it will try again until the number of attempts reaches retries, then drops the message.
+	Retries int `json:"retries,omitempty"`
+	// The time to wait in seconds before a dead connection is reestablished. (default: 60)
+	TimeReopen int `json:"time_reopen,omitempty"`
+	//
```
kristofgyuracz, line 79: Please document the field.

kristofgyuracz (changes requested):

pepov (approved):

kristofgyuracz (approved):
LGTM, thanks.

PR #1464, opened 2023-09-06
chore(syslog-ng): limit maxConnections and tune the node based increments

pepov:


OverOrion (approved):
LGTM, thanks for the quick fix!

PR #1481, opened 2023-09-18
health-check and storage-metrics support for fluent-bit

### LINKED ISSUE #1286: Add support to enable health-check and storage-metrics for fluent-bit in loggings-file
Hi their, so I tell you what's about . . . 

We're using logging-operator for quiet sometime and lately sometimes we notice fluentd is not receiving any logs from fluent-bit pods without a noticable reason. Redeploying is fixing it, but liveness- and readiness probes could handle this without any user interaction. 

Fluent-bit can provide health information and additional storage metrics to monitor. Both can be enabled in fluent-bit.conf, but logging-operator reverts changes in the corresponding secret as we save it. I've done some test with logging-operator scaled down.

I'd like to enable health-check and storage-metrics for fluent-bit via logging-operators logging-file. And then logging-operator sets "Health_Check" and "storage.metrics" in the secret that's mounted as fluent-bit.conf.

There seem's to be not really an alternative to get similiar metrics from the fluent-bits managed by logging-operator I can think of.

**Additional context**
- The health-check in fluent-bit I've mentioned is described here: https://docs.fluentbit.io/manual/administration/monitoring#health-check-for-fluent-bit
- The storage-metrics option is described here at the end of the table: https://docs.fluentbit.io/manual/administration/buffering-and-storage#service-section-configuration

nak0f:
PR for adding support to enable health-check and storage-metrics for fluent-bit. (fixes #1286)


OverOrion:
Thank you for your contribution! The CI fails because there are generated files which had not been pushed. Could you please add them as well? `make check-diff` should tell you what the difference is, currently it seems that `docs/configuration/crds/v1beta1/fluentbit_types.md` is the culprit.

nak0f:
> Thank you for your contribution! The CI fails because there are generated files which had not been pushed. Could you please add them as well? `make check-diff` should tell you what the difference is, currently it seems that `docs/configuration/crds/v1beta1/fluentbit_types.md` is the culprit.

Thanks @OverOrion :) That should be sorted out now.

pkg/resources/fluentbit/config.go
```diff
@@ -42,6 +42,22 @@ var fluentBitConfigTemplate = `
     {{ $key }}  {{$value}}
     {{- end }}
     {{- end }}
+    {{- if .HealthCheck }}
+    {{- if .HealthCheck.Enabled }}
+    Health_Check On
+    {{- else }}
+    Health_Check Off
+    {{- end }}
```
pepov, line 50: does it make sense to have a specific flag for this? wouldn't we want to enable healthcheck always if the healthcheck struct is defined?

nak0f: I just replicated the same config structure from fluent-bit but fair point. From a logging-operator user perspective this flag is not really needed.

pepov (approved):

OverOrion (approved):
LGTM, thank you once again for your contribution!

PR #1483, opened 2023-09-19
chore: global generate makefile rule

aslafy-z:
Rename the `generate` makefile rule to `codegen`
Add a new `generate` rule that runs `codegen fmt manifests docs helm-docs`
`check-diff` rule now calls `generate`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,11 @@ all: manager
 .PHONY: check
 check: license-check lint test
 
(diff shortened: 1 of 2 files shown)


pepov (approved):

pepov (approved):

OverOrion (approved):
Thanks for this, LGTM!

PR #1489, opened 2023-09-26
fix(fluentbit): increase default initial delay seconds

aslafy-z:
When fluent-bit wal is large, fluentbit can take some time to get back on track.


aslafy-z:
I think the liveness probe should change to use the health endpoint instead of the prometheus metrics one, which is not exposed at the time fluentbit register tails:
```
[2023/09/26 16:39:04] [ info] [input:storage_backlog:storage_backlog.1] register tail.0/1-1695742686.611226256.flb
[2023/09/26 16:39:04] [ info] [input:storage_backlog:storage_backlog.1] register tail.0/1-1695742686.615852633.flb
[2023/09/26 16:39:04] [ info] [input:storage_backlog:storage_backlog.1] register tail.0/1-1695742686.630729962.flb
```

PR #1491, opened 2023-09-27
docs: Add description for multi-worker fluentd

### LINKED ISSUE #1421: docs: multi worker guide
Guide on how to setup multi worker 

- set `fluentdSpec.workers`
- set `fluentdSpec.rootDir: '/buffers'`
- use `fluentdSpec.bufferStorageVolume.pvc` storage backend
- more to add?

florianstoeber:
I am adding a walkthrough of how to enable multi-worker fluentd.
The setup is opinionated and might be more generic, but at least it helped solve my problems.

Closes #1421


aslafy-z:
That's nice! Here some more things to add: https://github.com/kube-logging/logging-operator/issues/1421. 
If you integrate those changes, you can mark this PR as resolving #1421. I'm currently testing a multiprocess setup and may have some more to add in the few next days too.

florianstoeber:
Thank you for the suggestion @aslafy-z . I added the configs that are mentioned in the issue and marked the issue to be solved by this PR.

pepov:
I'm also thinking we should add the rootDir="/buffers" automatically if workers are configured, I'm not sure why we don't do that already...

docs/multi-worker.md
```diff
@@ -0,0 +1,133 @@
+# Mutli-Worker Fluentd Setup
+
+## Necessity
+
+In specific scenarios, a fluentd with a single worker instance cannot process and forward the high amount of logs produced on clusters. This can lead to fluentd Pods not accepting additional traffic from fluent-bits and fluent-bits suffering under Backpressure. In the end, both fluentd and fluent-bit Pods might run into their memory limits and get restarted by Kubernetes. Enabling multiple worker processes per fluentd Pod will increase the performance of this component, so it is recommended to use a multi-worker approach in environments with high log volume. Additionally the official [fluentd documentation](https://docs.fluentd.org/deployment/multi-process-workers) might be helpful.
+
+## Recommended implementation
+
+When enabling the multi-worker setup, it is recommended to ensure the following things:
+- Place the fluentds on separate Nodes
```
pepov, line 10: nit: I beleive that should be achived by antiaffinity rules, we can mention it or add to the example as well (optional)

docs/multi-worker.md
```diff
@@ -0,0 +1,133 @@
...
+- Increase the compute and memory resources 
+- Configure a buffer volume for fluentd
```
pepov, line 12: nit: fluentd is configured with a buffer volume by default, I'm not entirely sure this is required (but at least it doesn't hurt either)

docs/multi-worker.md
```diff
@@ -0,0 +1,133 @@
...
+    type: cpu
+  name: node1
+spec:
+  taints:
+  - effect: NoSchedule
+    key: type
+    value: cpu
+```
+
+The corresponding setting in the Logging-CRD is looking like follows:
```
pepov, line 32:
```suggestion
The corresponding setting in the Logging resource looks like follows:
```

docs/multi-worker.md
```diff
@@ -0,0 +1,133 @@
...
+```yaml
+- Limits:
+  - cpu: 1000m
+  - memory: 400M
+- Requests:
+  - cpu: 500m
+  - memory:  100M
+```
+
+In this short walkthrough, we will increase the fluentd workers from `1` to `5`. Therefore, we will multiply the requests and limits with factor 5 to ensure enough resources are reserved. Additionally, we will set requests and limits to the same values to ensure that the fluentd Pods are not affected by other workloads on the Node. This is, in general, a good practice. We do this by changing the Logging-CRD like follows:
```
pepov, line 54:
```suggestion
In this short walkthrough, we will increase the fluentd workers from `1` to `5`. Therefore, we will multiply the requests and limits with factor 5 to ensure enough resources are reserved. Additionally, we will set requests and limits to the same values to ensure that the fluentd Pods are not affected by other workloads on the Node. This is, in general, a good practice. We do this by changing the Logging resource as follows:
```

docs/multi-worker.md
```diff
@@ -0,0 +1,133 @@
...
+```yaml
+fluentd:
+  nodeSelector:
+    type: cpu
+  tolerations:
+  - effect: NoSchedule
+    key: type
+    operator: Equal
+    value: cpu
```
pepov, line 63:
```suggestion
```
I would only highlight the relevant part in each step as you show the complete config in the last step anyways

docs/multi-worker.md
```diff
@@ -0,0 +1,133 @@
...
+  resources:
+    limits:
+      cpu: 5
+      memory: 2G
+    requests:
+      cpu: 5
+      memory: 2G
+  rootDir: /buffers
```
pepov, line 91: This setting should be changed together with the workers setting, I would move it there and highlight these should be set together.

docs/multi-worker.md
florianstoeber: This was not my intention. I wanted to have separate Nodes for the fluentds, so that other workload is not scheduled on the same Node.

docs/multi-worker.md
florianstoeber: I will change that

docs/multi-worker.md
florianstoeber: I will change that

florianstoeber: changed

docs/multi-worker.md
florianstoeber: Changed

docs/multi-worker.md
pepov: fair enough

docs/multi-worker.md
florianstoeber: You are right, I will remove it.

pepov (approved):

PR #1494, opened 2023-09-28
fix(webhook-handler): followup regarding decoding changes in controller-runtime

OverOrion:
Also added some logs / error reporting where it seemed sensible.

Fixes #1493.

PR #1503, opened 2023-09-29
Make mutating webhook idempotent

### LINKED ISSUE #1444: Tailer-Webhook Producing an Invalid Spec

**Describe the bug**:
After setting up the Tailer-Webhook according to the [Docs](https://kube-logging.dev/docs/configuration/extensions/tailer-webhook/#multi-container-pods), it does not seem generate a valid spec and instead produces one with duplicate volume name, mountPath, and container name. It is as though either 2 mutations occurred or the operator tried to mutate twice. I have confirmed that there is only a single operator pod, service, and  mutating webhook.

Test:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  annotations:
    "sidecar.logging-extensions.banzaicloud.io/tail": "/var/log/date"
spec:
  containers:
    - image: debian
      name: sample-container
      command: ["/bin/sh", "-c"]
      args:
        - while true; do
          date >> /yolo/date;
          sleep 1;
          done
```

Result:
```bash
The Pod "test-pod" is invalid:
* spec.volumes[2].name: Duplicate value: "sample-container-var-log"
* spec.containers[0].volumeMounts[2].mountPath: Invalid value: "/var/log": must be unique
* spec.containers[2].name: Duplicate value: "sample-container-var-log-date"
```

**Expected behaviour**:
A concise description of what you expected to happen.

1. Setup Prerequisites per above Docs(Cert)
2. Setup Webhook(ENV var, MutatingWebhookConfig, Service)
3. Attempt to create the example pod.

**Additional context**:
We tested this across 4 identical  clusters, all with the same results.

**Environment details**:
- Kubernetes version (e.g. v1.15.2): 1.27.4
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): EKS
- logging-operator version (e.g. 2.1.1): 4.2.3
- Install method (e.g. helm or static manifests): [Operator Helm Chart](https://github.com/kube-logging/helm-charts/tree/main/charts/logging-operator) and [Logging Helm Chart](https://github.com/kube-logging/helm-charts/tree/main/charts/logging-operator-logging)

/kind bug

OverOrion:
Needs https://github.com/kube-logging/logging-operator/pull/1494 before merging.

Fixes #1444.

### DIFF
--- a/pkg/webhook/podhandler/podhandler.go
+++ b/pkg/webhook/podhandler/podhandler.go
@@ -137,7 +137,7 @@ func (p *PodHandler) podHandlerHelper(podToModify *corev1.Pod, sideCars []corev1
 			return c.Name == sideCar.Name
 		}) {
 			p.Log.Info(duplicateValuesMsg)
-			rv := admission.Allowed(duplicateValuesMsg)
+			rv := admission.Denied(duplicateValuesMsg)
 			return &rv
 
 		} else {
@@ -153,7 +153,7 @@ func (p *PodHandler) podHandlerHelper(podToModify *corev1.Pod, sideCars []corev1
 			return v.Name == volume.Name
 		}) {
 			p.Log.Info(duplicateValuesMsg)
-			rv := admission.Allowed(duplicateValuesMsg)
+			rv := admission.Denied(duplicateValuesMsg)
 			return &rv
 		} else {
 			podToModify.Spec.Volumes = append(podToModify.Spec.Volumes, volume)
--- a/pkg/webhook/podhandler/podhandler.go
+++ b/pkg/webhook/podhandler/podhandler.go
@@ -126,8 +126,7 @@ func (p *PodHandler) Handle(ctx context.Context, req admission.Request) admissio
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 
-	rv := admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
-	return rv
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
 
 func (p *PodHandler) podHandlerHelper(podToModify *corev1.Pod, sideCars []corev1.Container, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount) *admission.Response {
--- a/pkg/webhook/podhandler/podhandler.go
+++ b/pkg/webhook/podhandler/podhandler.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kube-logging/logging-operator/pkg/resources/kubetool"
 	"github.com/kube-logging/logging-operator/pkg/resources/volumepath"
 	config "github.com/kube-logging/logging-operator/pkg/sdk/extensions/extensionsconfig"
+	"github.com/siliconbrain/go-seqs/seqs"
 	corev1 "k8s.io/api/core/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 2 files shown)


pkg/webhook/podhandler/podhandler.go
```diff
@@ -125,5 +126,40 @@ func (p *PodHandler) Handle(ctx context.Context, req admission.Request) admissio
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	rv := admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	return rv
```
pepov, line 130:
```suggestion
	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
```

pkg/webhook/podhandler/podhandler.go
```diff
@@ -108,15 +109,15 @@ func (p *PodHandler) Handle(ctx context.Context, req admission.Request) admissio
...
 
 		sideCars, volumes, volumeMounts := p.sideCarsForContainer(container.Name, filePaths)
 
 		// Append the new data to the podspec
-		pod.Spec.Containers = append(pod.Spec.Containers, sideCars...)
-		pod.Spec.Volumes = append(pod.Spec.Volumes, volumes...)
-		pod.Spec.Containers[idx].VolumeMounts = append(pod.Spec.Containers[idx].VolumeMounts, volumeMounts...)
+		if err := p.podHandlerHelper(pod, sideCars, volumes, volumeMounts); err != nil {
+			return *err
+		}
```
pepov, line 120:
```suggestion
		if resp := p.podHandlerHelper(pod, sideCars, volumes, volumeMounts); resp != nil {
			return *resp
		}
```

pkg/webhook/podhandler/podhandler.go
```diff
@@ -125,5 +126,40 @@ func (p *PodHandler) Handle(ctx context.Context, req admission.Request) admissio
...
+
+func (p *PodHandler) podHandlerHelper(podToModify *corev1.Pod, sideCars []corev1.Container, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount) *admission.Response {
+	duplicateValuesMsg := "webhook mutation would result in duplicate values, returning"
+	for idx, sideCar := range sideCars {
+		sideCar := sideCar
+		if seqs.Any(seqs.FromSlice(podToModify.Spec.Containers), func(c corev1.Container) bool {
+			return c.Name == sideCar.Name
+		}) {
+			p.Log.Info(duplicateValuesMsg)
+			rv := admission.Allowed(duplicateValuesMsg)
```
pepov, line 141: Why not admission.Denied? I beleive this should result in an error on the user side

pkg/webhook/podhandler/podhandler.go
```diff
@@ -125,5 +126,40 @@ func (p *PodHandler) Handle(ctx context.Context, req admission.Request) admissio
...
+		}
+	}
+
+	for _, volume := range volumes {
+		volume := volume
+		if seqs.Any(seqs.FromSlice(podToModify.Spec.Volumes), func(v corev1.Volume) bool {
+			return v.Name == volume.Name
+		}) {
+			p.Log.Info(duplicateValuesMsg)
+			rv := admission.Allowed(duplicateValuesMsg)
```
pepov, line 157: same here: I would consider denying the request here

pkg/webhook/podhandler/podhandler.go
OverOrion: My reasoning was that it would not be idempotent if we denied the request, but after giving it a bit more thought it seems to be idempotent on the server side (appends only if needed) but the user indeed should receive an error.

aslafy-z:
Will this be ported to main?

PR #1535, opened 2023-10-13
logging route watch

pepov:
- fix(webhook-handler): followup regarding decoding changes in controller-runtime
- fix(webhook): make the handler idempotent
- simplify fluentd securitycontext propagation and apply to the buffer metrics sidecar as well
- securitycontext and resources for the drainer sidecar
- securitycontext for the fluentd compressed config init container
- Update pkg/webhook/podhandler/podhandler.go
- Update pkg/webhook/podhandler/podhandler.go
- webhook: deny the user's request if it would not cause any changes
- fix(loggingroute): add missing watch for the loggingroute controller


OverOrion (approved):
LGTM, thanks for the detailed explanation!

Issue #1495, opened 2023-09-28
Resource CLI output and status updates


pepov:
extend the status and cli output with
- [x] logging watchnamespaces 
- [x] add logging configcheck failure message into logging problems

Follow up items: https://github.com/kube-logging/logging-operator/issues/1544

PR #1542, opened 2023-10-16
Updates in logging status problems for usability

pepov:
Usability updates to highlight certain issues in the logging resource status:
- deprecated embedded fluentbit config is used together with logging routes
- watchNamespaceSelector doesn't resolve to actual namespaces

#1495

### DIFF
--- a/pkg/resources/model/repository.go
+++ b/pkg/resources/model/repository.go
@@ -69,6 +69,9 @@ func (r LoggingResourceRepository) LoggingResourcesFor(ctx context.Context, logg
 	res.Fluentbits, err = r.FluentbitsFor(ctx, logging)
 	errs = errors.Append(errs, err)
 
+	res.LoggingRoutes, err = r.LoggingRoutesFor(ctx, logging)
+	errs = errors.Append(errs, err)
+
 	res.WatchNamespaces, err = UniqueWatchNamespaces(ctx, r.Client, &logging)
 	if err != nil {
 		errs = errors.Append(errs, err)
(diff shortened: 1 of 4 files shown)


OverOrion (approved):

pkg/resources/model/repository.go
```diff
@@ -334,6 +337,25 @@ func (r LoggingResourceRepository) FluentbitsFor(ctx context.Context, logging v1
...
+	sort.Slice(list.Items, func(i, j int) bool {
+		return lessByNamespacedName(&list.Items[i], &list.Items[j])
+	})
+
+	var res []v1beta1.LoggingRoute
+	for _, i := range list.Items {
+		if i.Spec.Source == logging.Spec.LoggingRef {
+			res = append(res, i)
+		}
+	}
```
siliconbrain, line 355: sorting could be done _after_ filtering so only the "important" items are sorted
```suggestion
	var res []v1beta1.LoggingRoute
	for _, i := range list.Items {
		if i.Spec.Source == logging.Spec.LoggingRef {
			res = append(res, i)
		}
	}

	sort.Slice(res, func(i, j int) bool {
		return lessByNamespacedName(&res[i], &res[j])
	})
```

siliconbrain (approved):
lgtm

pkg/resources/model/repository.go
pepov: @siliconbrain changed for all the occurences in b8bdce02

pepov: sorry, its 7ca6841a

PR #1561, opened 2023-10-27
fix(outputs): use simple strings as parameters

kristofgyuracz:


siliconbrain (approved):
lgtm

pkg/sdk/logging/model/syslogng/output/mongodb.go
```diff
@@ -76,6 +78,7 @@ type MongoDB struct {
 	TimeReopen int `json:"time_reopen,omitempty"`
 	// 	Description: Sets the write concern mode of the MongoDB operations, for both bulk and single mode.
 	// See [syslog-ng docs] https://axoflow.com/docs/axosyslog-core/chapter-destinations/configuring-destinations-mongodb/reference-destination-mongodb/#mongodb-option-write-concern
+	// +kubebuilder:validation:Enum=unacked;acked;majority
```
OverOrion, line 81: With this validation added, the option to set a custom integer as a value for the `write_concern` is not possible. How about removing that test case?

OverOrion (approved):
LGTM, thanks for it!

PR #1576, opened 2023-11-06
fix(charts/logging-operator): Improve tailer (image, docs and helm cleanup)

genofire:
- Support image
- Move default values into `values.yaml` (and add BreakingChange `logging.eventTailer.enabled` as false by default)
- Move default values of pvc to `values.yaml` (and add BreakingChange `logging.eventTailer.pvc.enabled` as false by default)
- Move default values of hostTailers as well and also add `enabled` flag as false by default


charts/logging-operator/values.yaml
```diff
@@ -239,15 +239,38 @@ logging:
...
-    # name: sample
-    # pvc:
-    #   accessModes:
-    #     - ReadWriteOnce
-    #   volumeMode: Filesystem
-  #   storage: 1Gi
-  #   storageClassName: standard
+  # EventTailer config
+  eventTailer:
+    create: false
```
pepov, line 244: nit: I would prefer enabled to be consistent

pepov (approved):
looks great, thanks!

charts/logging-operator/values.yaml
genofire: done

genofire:
i move also all logging parts into a directory

pepov:
If we disable eventTailer shouldn't we disable hosttailers as well by default? Could you perhaps add that as well?

genofire:
okay, ready to review again

charts/logging-operator/values.yaml
```diff
@@ -239,28 +239,78 @@ logging:
...
+      repository:
+      # -- tag of eventTailer image
+      tag:
+      # -- pullPolicy of eventTailer image
+      pullPolicy:
+      # -- imagePullSecrets of eventTailer image
+      imagePullSecrets:
+    pvc:
+      # -- enable pvc for 
+      enabled: false
```
pepov, line 257: I would recommend to keep this enabled by default to keep compatibility

pepov:
Looks good to me, have you been able to test it?

genofire:
not yet - and i am not sure if i could spend time this year. For us it is an requirement for an update, so i just make it as prepeare it.

genofire:
any news?

charts/logging-operator/values.yaml
```diff
@@ -226,28 +226,78 @@ logging:
...
+    name: event-tailer
+    image:
+      # -- repository of eventTailer image
+      repository:
+      # -- tag of eventTailer image
+      tag:
+      # -- pullPolicy of eventTailer image
+      pullPolicy:
+      # -- imagePullSecrets of eventTailer image
+      imagePullSecrets:
```
OverOrion, line 241: I think `imagePullSecrets` type should be `list` not `string`.

It applies to other places as well.

OverOrion (dismissed):

charts/logging-operator/values.yaml
genofire: you mean the default should be that way?
```suggestion
      imagePullSecrets: []
```

OverOrion: Yes, currently the generated README tells it's a string, because the default value is not explicitly defined.

genofire:
please reopen it -> otherwise i could not work on it.

genofire:
done and hopefull rebased correct after merge of https://github.com/kube-logging/logging-operator/pull/1674

pepov:
I've tested, rebased and fixed indentation issues, good to go from my side. @OverOrion please take a look!

kristofgyuracz (approved):

genofire:
oh thank you for doing my work _duck_

Issue #1565, opened 2023-10-30
Support custom blob suffix for Azure (e.g gov cloud)


mjnovice:
**Is your feature request related to a problem? Please describe.**
I am wanting to use this operator on Azure Government cloud, which have a custom storage blob suffix `core.usgovcloudapi.net` , unlike what the container `fluentd-1.15-ruby3` defaults to  `core.windows.net` 

**Describe the solution you'd like**
Allow one to specify the storage blob suffix.

**Describe alternatives you've considered**
Not sure if there is an alternative ?

**Additional context**

mjnovice:
cc - @tarokkk @ahma Could you share thoughts 🙏

PR #1579, opened 2023-11-09
Adding AzureCloud field optionally usable by plugins

mjnovice:
Attempts at solving #1565

### DIFF
--- a/pkg/sdk/logging/model/output/azurestore.go
+++ b/pkg/sdk/logging/model/output/azurestore.go
@@ -40,6 +40,9 @@ type _metaAzure interface{} //nolint:deadcode,unused
 type AzureStorage struct {
 	// Path prefix of the files on Azure
 	Path string `json:"path,omitempty"`
+	// Azure Cloud to use. e.g. AzurePublicCloud, AzureChinaCloud, AzureGermanCloud, AzureUSGovernmentCloud
+	// This field is supported only if the fluentd plugin honors it. e.g https://github.com/elsesiy/fluent-plugin-azure-storage-append-blob-lts
+	AzureCloud string `json:"azure_cloud,omitempty"`
 	// Your azure storage account
 	// +docLink:"Secret,../secret/"
 	AzureStorageAccount *secret.Secret `json:"azure_storage_account"`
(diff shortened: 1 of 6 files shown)


OverOrion:
Hi @mjnovice, 

Thank you for your contribution! Currently the CI fails, because the sign-off line is missing from the commit. Could you please add it?

See the [DCO check](https://github.com/kube-logging/logging-operator/pull/1579/checks?check_run_id=18525932298) for details.

pepov (approved):

pkg/sdk/logging/model/output/azurestore.go
```diff
@@ -40,6 +40,9 @@ type _metaAzure interface{} //nolint:deadcode,unused
 type AzureStorage struct {
 	// Path prefix of the files on Azure
 	Path string `json:"path,omitempty"`
+	// Azure Cloud to use. e.g. AzurePublicCloud, AzureChinaCloud, AzureGermanCloud, AzureUSGovernmentCloud
+	// This field is supported only if the fluentd plugin honors it. e.g https://github.com/elsesiy/fluent-plugin-azure-storage-append-blob-lts
+	AzureCloud string `json:"azure_cloud,omitempty"`
```
aslafy-z, line 45: If we want to support Azure Stack Cloud, we may need to precise it should be in uppercase: `AZURESTACKCLOUD`

mjnovice: Hey @aslafy-z , thanks for the comment. Do you mean adding it as an example in the comment ?

pkg/sdk/logging/model/output/azurestore.go
```diff
@@ -40,6 +40,9 @@ type _metaAzure interface{} //nolint:deadcode,unused
 type AzureStorage struct {
 	// Path prefix of the files on Azure
 	Path string `json:"path,omitempty"`
+	// Azure Cloud to use. e.g. AzurePublicCloud, AzureChinaCloud, AzureGermanCloud, AzureUSGovernmentCloud
```
aslafy-z, line 43:
```suggestion
	// Azure Cloud to use. e.g. AzurePublicCloud, AzureChinaCloud, AzureGermanCloud, AzureUSGovernmentCloud, AZURESTACKCLOUD (keep uppercase).
```
Something like this would do the thing I think :)

pkg/sdk/logging/model/output/azurestore.go
aslafy-z: I made a proposal :)

aslafy-z:
Please regenerate the docs with `make docs` 🙏

aslafy-z (approved):

Issue #1477, opened 2023-09-16
Support AWS credentials authentication for OpenSearch output


cyrus-mc:
**Is your feature request related to a problem? Please describe.**

When running OpenSearch via AWS OpenSearch service you should be able to authenticate using AWS credentials as documented in the plugin documentation (https://github.com/fluent/fluent-plugin-opensearch).

**Describe the solution you'd like**

Add the following options to the OpenSearch output:

* access_key_id
* secret_access_key
* assume_role_arn
* assume_role_session_name
* sts_credentials_region
* region
* assume_role_arn
* assume_role_web_identity_token_file

**Describe alternatives you've considered**

Work around is to configure open access to OpenSearch from within account VPC.

PR #1582, opened 2023-11-16
Adding Opensearch field endpoint to support AWS Opensearch

jwitrick:
This PR addresses the missing ability to use the logging-operator with a secure AWS Opensearch cluster.

Solves this issue: https://github.com/kube-logging/logging-operator/issues/1477


pkg/sdk/logging/model/output/opensearch.go
```diff
@@ -278,3 +294,34 @@ func (e *OpenSearchOutput) ToDirective(secretLoader secret.SecretLoader, id stri
...
 }
+
+
+
+type OpenSearchEndpointCredentials struct {
+	// AWS region. It should be in form like us-east-1, us-west-2. Default nil, which means try to find from environment variable AWS_REGION.
+	Region string `json:"region,omitempty"`
+
+	// AWS connection url.
+	Url string `json:"url,omitempty"`
```
pepov, line 305: nit: shouldn't this be mandatory?

pepov (approved):
looks wonderful, thanks!

OverOrion (commented):
Thank you for your contribution @jwitrick! Currently the [CI fails ](https://github.com/kube-logging/logging-operator/actions/runs/6894811966/job/18775701005?pr=1582)because some generated code has not been committed.
Could you run `make check-diff` and commit the newly generated files?

pkg/sdk/logging/model/output/opensearch.go
jwitrick: Looking at the plugin i think you are correct it should be. Fixing now.

jwitrick: Changes pushed. Please take another look.

PR #1583, opened 2023-11-20
fix(reconciler): make sure to not update the original object

OverOrion:
If the original object contains a pointer, then without this patch that pointer will (inadvertently) change.

This is required for the detached fluentd use case.

### DIFF
--- a/pkg/resources/model/reconciler.go
+++ b/pkg/resources/model/reconciler.go
@@ -277,7 +277,8 @@ func NewValidationReconciler(
 				continue
 			}
 
-			if err := repo.Status().Patch(ctx, req.Obj, req.Patch); err != nil {
+			obj := req.Obj.DeepCopyObject().(client.Object) // copy object so that the original is not changed by the call to Patch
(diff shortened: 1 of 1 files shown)


pepov (approved):

pkg/resources/model/reconciler.go
```diff
@@ -276,8 +276,9 @@ func NewValidationReconciler(
 			if req.IsEmptyPatch() {
 				continue
 			}
+			obj := req.Obj.DeepCopyObject()
```
siliconbrain, line 279: could address the reason for the copy with a comment
```suggestion
			obj := req.Obj.DeepCopyObject() // copy object so that the original is not changed by the call to Patch
```

siliconbrain, line 279: could move the type assertion here, where it's more obviously valid (since `req.Obj` is a `client.Object` by definition, thus the copy will be too...)
```suggestion
			obj := req.Obj.DeepCopyObject().(client.Object)
```

siliconbrain (approved):

PR #1587, opened 2023-11-21
feat(syslogng-http-outputs): add ssl_version field for all HTTP based outputs

kristofgyuracz:
Validations added for input, can be omitted, if we want to accept all upper-lowercase combinations.


OverOrion (approved):
LGTM, thanks for this!

PR #1589, opened 2023-11-21
feat(openobserve): add syslog-ng's openobserve output

OverOrion:
TODOs:

- [x] Add example (optional)
- [x] Followup changes (-log suffix, record)


pkg/sdk/logging/api/v1beta1/syslogng_output_types.go
```diff
@@ -45,6 +45,7 @@ type SyslogNGOutputSpec struct {
 	SplunkHEC       *output.SplunkHECOutput       `json:"splunk_hec_event,omitempty" syslog-ng:"dest-drv"`
 	Loki            *output.LokiOutput            `json:"loki,omitempty" syslog-ng:"dest-drv"`
 	S3              *output.S3Output              `json:"s3,omitempty" syslog-ng:"dest-drv"`
+	Openobserve     *output.OpenobserveOutput     `json:"openobserve-log,omitempty" syslog-ng:"dest-drv"`
```
pepov, line 48: can't we use openobserve as the key? I would rather name this openobserve and if we happen to want to send something else (like metrics) then we can name it openobserver-metrics, but I don't see that happening at the moment

pkg/sdk/logging/model/syslogng/output/openobserve.go
```diff
@@ -0,0 +1,74 @@
...
+    organization: "default"
+    stream: "default"
+    user: "username"
+    password:
+      valueFrom:
+        secretKeyRef:
+          name: openobserve
+          key: password
+{{</ highlight >}}
+More information at TODO
```
pepov, line 43:
```suggestion
More information at https://axoflow.com/docs/axosyslog-core/chapter-destinations/openobserve/
```

pkg/sdk/logging/model/syslogng/output/openobserve.go
```diff
@@ -0,0 +1,74 @@
...
+*/
+type _docOpenobserve interface{} //nolint:deadcode,unused
+
+// +name:"Openobserve"
+// +url:"TODO"
```
pepov, line 48:
```suggestion
// +url:"https://axoflow.com/docs/axosyslog-core/chapter-destinations/openobserve/"
```

pkg/sdk/logging/api/v1beta1/syslogng_output_types.go
OverOrion: Renamed it accordingly.

PR #1594, opened 2023-11-26
feat: add support for useragent Fluent filter

csabatuz-chess:
This PR adds support for using the https://github.com/bungoume/fluent-plugin-ua-parser Fluent plugin where Fluent plugins are applicable.

Depends on: https://github.com/kube-logging/fluentd-images/pull/88


pkg/sdk/logging/model/filter/useragent.go
```diff
@@ -0,0 +1,103 @@
...
+//
+//		@type ua_parser
+//		@id test_useragent
+//	 key_name my_agent
+//	 delete_key true
+//	 out_key ua_fields
+//	 flatten true
+//
+// </filter>
+// ```
```
pepov, line 84: please follow multiline formatting pattern based on this example:
https://github.com/kube-logging/logging-operator/blob/b4191ebb4d856b64cc5b6600d02e22c357b4a586/pkg/sdk/logging/model/syslogng/output/s3.go#L25

csabatuz-chess: If yes, would you like me to do that for all the filters in a separate PR?

pepov: It seems we haven't fixed all the bad patterns yet. You don't have to fix all the other filters, but let's format this one properly.

Follow up: https://github.com/kube-logging/logging-operator/issues/1596

csabatuz-chess: Ok, I don't know if what I did is what you meant, but I did the best I could figure.

csabatuz-chess:
I'm stuck a bit, can't really interpret the failing E2E test, or connect it to my changeset.

OverOrion:
> I'm stuck a bit, can't really interpret the failing E2E test, or connect it to my changeset.

I reran the CI and it passed, seems like the `TestVolumeDrain_Downscale` might be flaky, sorry about that!

pkg/sdk/logging/model/filter/useragent.go
pepov: thanks, I've made a few updates to the formatting, but seems to be good now!

pepov (approved):

csabatuz-chess:
image
The <highlight xml> directives don't seem to render right in Markdown, that's why I left with ```yaml.

Not sure what's downstream, if this'll be straightened out in the docs, just checking in if this was intended.

OverOrion:
> image The directives don't seem to render right in Markdown, that's why I left with ```yaml.
> Not sure what's downstream, if this'll be straightened out in the docs, just checking in if this was intended.

Yeah, that is alright, it is deliberate. The generated *.md files will be consumed by Hugo and it will render it properly (indentation and highlighting wise as well).

OverOrion (approved):

Issue #1520, opened 2023-10-10
PodDisruptionBudget for fluentd


florianstoeber:
**Is your feature request related to a problem? Please describe.**
In maintenance situations (Kubernetes update, node-image update, etc.), too many fluentds may be drained from the nodes. This can lead to an overflow situation in the fluentds, and the stability of the logging system is not guaranteed.   

**Describe the solution you'd like**
The FluentdSpec should be enhanced to have a pdb field. This can be configured optionally in the following regard and create a corresponding PDB for the fluentd Statefulset:

```yaml
fluentd:
  pdb:
    minAvailable: 2
```

```yaml
fluentd:
  pdb:
    minUnavailable: 10%
```

**Describe alternatives you've considered**
An alternative for me is to create the PDB manually or with an umbrella helm-chart. There might also be other solutions, but creating a PDB for the component is the easiest and Kubernetes-native way to guarantee way availability of the fluentd Pods.

**Additional context**
https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget

pepov:
this has been fixed in #1602

PR #1602, opened 2023-12-02
Feat: add PodDisruptionBudget for fluentd

yunkunrao:
This pull request is related with #1520 and add a new pdb field to enhance FluentdSpec.


yunkunrao:
> Hey @yunkunrao , thanks for the PR!
> 
> It seems to me, that we let the user configure a full `pdb`, but then we overwrite the selector completely. I would suggest to let the user configure the fields that we actually use. What do you think?

You mean we let user themselves to create pdb and our logging controller override the pdb spec. selector field with new selector in fluentd. Well, it makes sense.

pepov:
@yunkunrao no, I meant that by using the poddisruptionbudget type instead of just minUnavailable, maxUnavailable and unhealthyPodEvictionPolicy the user might be confused whether what they should configure as selectors (which they don't have to touch)

yunkunrao:
> @yunkunrao no, I meant that by using the poddisruptionbudget type instead of just minUnavailable, maxUnavailable and unhealthyPodEvictionPolicy the user might be confused what they should configure for selectors (which they don't have to touch)

Got it. Maybe we should only expose just minUnavailable, maxUnavailable and unhealthyPodEvictionPolicy to the user for easy use and avoid misunderstanding.

pepov (approved):

OverOrion (approved):

pepov:
thanks @yunkunrao !

Issue #1616, opened 2023-12-11
Custom PVC source specified in ExtraVolumes in the fluentd definition is not honoured 


bbyreddy:
**Describe the bug**:
Custom PVC source details specified in the fluentd.extraVolumes.volume.pvc.source.claimName of the fluentd definition are not honoured by the logging-operator. It always tries to create a new PVC and expects a non-empty fluentd.extraVolumes.volume.pvc.spec.
Operator Version: 4.4.3

I suppose the issue is with the following code snippet in the pkg.resources.fluentd.statefulset.go (line# 51 to 63). The else block will never be executed as the n.Volume.PersistentVolumeClaim will not ne null in both the cases (when user wants to provide pvc.spec or pvc.source).

```
	for _, n := range r.Logging.Spec.FluentdSpec.ExtraVolumes {
		if n.Volume != nil && n.Volume.PersistentVolumeClaim != nil {
			if err := n.Volume.ApplyPVCForStatefulSet(n.ContainerName, n.Path, spec, func(name string) metav1.ObjectMeta {
				return r.FluentdObjectMeta(name, ComponentFluentd)
			}); err != nil {
				return nil, reconciler.StatePresent, err
			}
		} else {
			if err := n.ApplyVolumeForPodSpec(&spec.Template.Spec); err != nil {
				return nil, reconciler.StatePresent, err
			}
		}
	}
```

**Expected behaviour**:
Expectation according to the documentation is to either provide a volume.pvc.spec or volume.pvc.source. If the volume.pvc.source is provided, then the same should be mounted instead of creating a new PVC.

**Steps to reproduce the bug**:

1. Deploy the logging-operator and create the logging custom resource as defined below

```
apiVersion: logging.banzaicloud.io/v1beta1
kind: Logging
metadata:
  name: logging
spec:
  controlNamespace: c8yedge
  fluentbit:
    bufferStorageVolume:
      hostPath:
        path: ""
  fluentd:
    disablePvc: true
    bufferStorageVolume:
      emptyDir:
        medium: Memory
    extraArgs:
    - --log-rotate-age
    - "10"
    extraVolumes:
    - containerName: fluentd
      path: /var/log/
      volume:
        pvc:
          source:
            claimName: edge-logs
            readOnly: false
      volumeName: edge-logs
  watchNamespaces:
  - c8yedge
```

2. logging-fluentd statefulset fails to deploy with the following error

[...]

PR #1617, opened 2023-12-12
Enable using custom PVC without a template for the statefulset (for the buffer volume and also as extravolumes)

bbyreddy:
Ensured that the custom PVC source details specified in the fluentd.extraVolumes.volume.pvc.source.claimName of the fluentd definition are honoured instead of always attempting to create a new PVC.

Introduced empty checks to the PersistentVolumeClaim.PersistentVolumeClaimSpec and PersistentVolumeClaim.PersistentVolumeSource structures and ApplyPVCForStatefulSet() and ApplyVolumeForPodSpec() functions are invoked only when the respective structures are not empty.


pkg/resources/fluentd/statefulset.go
```diff
@@ -72,6 +77,16 @@ func (r *Reconciler) statefulset() (runtime.Object, reconciler.DesiredState, err
 	return desired, reconciler.StatePresent, nil
 }
 
+func isPersistentVolumeClaimSpecEmpty(pvcSpec corev1.PersistentVolumeClaimSpec) bool {
+	searilizedPvcSpec, _ := json.Marshal(pvcSpec) // Json serialization should not fail, hence the error is ignored
+	return string(searilizedPvcSpec) == `{"resources":{}}`
```
aslafy-z, line 82:
```suggestion
	serilizedPvcSpec, _ := json.Marshal(pvcSpec) // Json serialization should not fail, hence the error is ignored
	return string(serilizedPvcSpec) == `{"resources":{}}`
```

pkg/resources/fluentd/statefulset.go
```diff
@@ -72,6 +77,16 @@ func (r *Reconciler) statefulset() (runtime.Object, reconciler.DesiredState, err
...
+}
+
+func isPersistentVolumeSourceEmpty(pvcSource corev1.PersistentVolumeClaimVolumeSource) bool {
+	searilizedPvcSource, _ := json.Marshal(pvcSource) // Json serialization should not fail, hence the error is ignored
+	return string(searilizedPvcSource) == `{}`
```
aslafy-z, line 87:
```suggestion
	serilizedPvcSource, _ := json.Marshal(pvcSource) // Json serialization should not fail, hence the error is ignored
	return string(serilizedPvcSource) == `{}`
```

aslafy-z (changes requested):
I spot a couple of typos, can you please fix them?

pepov:
I promised to look into how we could make the comparison more efficient, sorry I haven't been able to get back to that

bbyreddy:
> I promised to look into how we could make the comparison more efficient, sorry I haven't been able to get back to that

Hi @pepov and @aslafy-z ,
There is another PR https://github.com/kube-logging/logging-operator/pull/1622 created by @OverOrion , which resolves this issue. I'm not sure which direction you want to take to resolve this issue.
Thanks!

pkg/resources/fluentd/statefulset.go
bbyreddy: Corrected the spelling.

pepov:
@bbyreddy we want to keep this instead of introducing yet another huge field to the fluentd spec

pepov:
I'm planning to add some technical docs as well

bbyreddy:
> @bbyreddy can you check if the solution that works for you?

@pepov , yes, this works perfectly fine and is much simpler as well. 
Please let me know if I need to do anything to get this PR merged into the master.
Thanks!

pepov (approved):

bbyreddy:
> I spotted a couple of typos, can you please fix them? Also, can you please update the PR name so it reflects the changes and add `fixes #1616` in the PR description? You will also have to sign your commits, see https://github.com/kube-logging/logging-operator/pull/1617/checks?check_run_id=20378816523.

@aslafy-z , you may have to mark this as resolved and approve the PR as the code has changed now and doesn't have the typos anymore. Thanks!

OverOrion (approved):

pepov:
@aslafy-z typos has been fixed and signoff done, I'm merging this

PR #1619, opened 2023-12-18
Generated doc fixes 4.5

fekete-robert:


pkg/sdk/logging/model/filter/concat.go
```diff
@@ -79,8 +79,8 @@ type Concat struct {
 	PartialCriStreamKey string `json:"partial_cri_stream_key,omitempty"`
 }
 
-// ## Example `Concat` filter configurations
 /*
+## Example `Concat` filter configurations
```
pepov, line 84: @fekete-robert this breaks go fmt, this is why the PR is failing

We could add an empty line comment if that works

fekete-robert: Oh :( Then it'll probably break at some other places as well, I thought this pattern will work

pepov: this breaks everywhere, just picked this as an example

fekete-robert: Ok, let's discuss IRL which pattern should we use, then I'll fix it everywhere

pepov (approved):

PR #1622, opened 2023-12-19
feat(fluentd): add externalVolumes field for externally managed volumes

### LINKED ISSUE #1616: Custom PVC source specified in ExtraVolumes in the fluentd definition is not honoured

OverOrion:
Instead of relying on the current implementation how the `extraVolumes` can be used to mount pre-existing volumes add a new `externalVolumes` field. This can be used to mount externally managed volumes. 

Fixes #1616.

### DIFF
--- a/charts/logging-operator/crds/logging.banzaicloud.io_fluentdconfigs.yaml
+++ b/charts/logging-operator/crds/logging.banzaicloud.io_fluentdconfigs.yaml
@@ -977,6 +977,179 @@ spec:
                   - name
                   type: object
                 type: array
+              externalVolumes:
+                items:
+                  properties:
+                    containerName:
+                      type: string
+                    path:
+                      type: string
+                    volume:
+                      properties:
+                        emptyDir:
+                          properties:
+                            medium:
+                              type: string
+                            sizeLimit:
+                              anyOf:
+                              - type: integer
+                              - type: string
+                              pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                              x-kubernetes-int-or-string: true
+                          type: object
+                        host_path:
+                          properties:
+                            path:
+                              type: string
+                            type:
+                              type: string
+                          required:
+                          - path
+                          type: object
+                        hostPath:
+                          properties:
+                            path:
+                              type: string
+                            type:
+                              type: string
+                          required:
+                          - path
+                          type: object
+                        pvc:
+                          properties:
+                            source:
+                              properties:
+                                claimName:
+                                  type: string
+                                readOnly:
+                                  type: boolean
+                              required:
+                              - claimName
+                              type: object
+                            spec:
+                              properties:
+                                accessModes:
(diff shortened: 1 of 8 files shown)


bbyreddy:
@OverOrion , these changes are working perfectly fine for my scenario, where I am creating a PVC and assigning the same to the external volume using this new field. 
Please go ahead with the merge.
Thanks!

pepov:
closing this in favour of keeping and extending #1617

PR #1656, opened 2024-01-30
fix: default parsers conf location wrong on windows

mweibel:
both linux and windows have it at the same location

Looking at that, I wonder why there are two different specs/types for the same thing (FluentbitSpec and NodeAgentFluentbit) - the former permits to override the config while the latter doesn't.

### DIFF
--- a/pkg/resources/nodeagent/config.go
+++ b/pkg/resources/nodeagent/config.go
@@ -23,7 +23,7 @@ var fluentBitConfigTemplate = `
     Grace        {{ .Grace }}
     Daemon       Off
     Log_Level    {{ .LogLevel }}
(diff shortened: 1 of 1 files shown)


pepov (approved):

OverOrion (approved):

mweibel:
@pepov @OverOrion hey! just wanted to ask about a release here - I didn't see the latest patches to include this change, any chance this gets into the next release? Thanks!

pepov:
@mweibel we don't automatically backport items, please open a pr against release-4.5 and we can merge/release from there as 4.5.x

mweibel:
ok, wasn't aware of that - opened #1663 for that

PR #1665, opened 2024-02-08
fix to make detached aggregators work when generating fluentbit tenants

pepov:


OverOrion (approved):
LGTM, thanks for the fix!

PR #1672, opened 2024-02-15
Fix fluentdconfig and syslogngconfig deletion

### LINKED ISSUE #1667: Fluentdconfig object deleted, pod is still running

**Describe the bug**:
Fluentdconfig object deleted but the pod is still running

**Expected behaviour**:
Fluentdconfig object deletion should stop the currently running pod.

Dispatch a fluentdconfig (or use the e2e test case), delete the fluentdconfig object and the pod is still there.

**Additional context**:
`logging.Status.FluentdConfigName` can be used to decide whether a fluentdconfig:
- does not exist YET (.Status.FluentdConfigName is not set)
- NO LONGER exists (.Status.FluentdConfigName is set, but object is not there)

Questions: 
- [ ] If a fluentdconfig gets deleted should be owning loggingspec be reusable for a different one?

Maybe weak references could help here?

**Environment details**:
- Kubernetes version (e.g. v1.15.2):
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc):
- logging-operator version (e.g. 2.1.1):
- Install method (e.g. helm or static manifests):

/kind bug

OverOrion:
This PR aims to fix two problems:
- deleting a fluentdconfig/syslogngconfig resource should also delete the child resources (solved by changing the ownerreferences)
- deleting a logging resource should not be possible until a fluentdconfig/syslogngconfig resource exists. we don't delete the fluentdconfig/syslogngconfig forcefully because those are created by the user, but we can protect against this by using a finalizer. we send an event to the logging object for the user to be able to see other than looking at the operator logs

Fixes #1667.

Signed-off-by: Szilard Parrag <szilard.parrag@axoflow.com>

### DIFF
--- a/controllers/logging/logging_controller.go
+++ b/controllers/logging/logging_controller.go
@@ -353,14 +353,14 @@ func (r *LoggingReconciler) checkSyslogNGConfigFinalizer(ctx context.Context, lo
 }
 
 func (r *LoggingReconciler) cleanupFluentdConfigReference(ctx context.Context, logging *loggingv1beta1.Logging, log logr.Logger) (ctrl.Result, error) {
-	log.Info("cleaned up fluentdConfigRef", "name", logging.Status.FluentdConfigName)
+	log.Info("cleaning up fluentdConfig reference from status", "name", logging.Status.FluentdConfigName)
 	logging.Status.FluentdConfigName = ""
 	err := r.Status().Update(ctx, logging)
 
 	return ctrl.Result{}, err
 }
 func (r *LoggingReconciler) cleanupSyslogNGConfigReference(ctx context.Context, logging *loggingv1beta1.Logging, log logr.Logger) (ctrl.Result, error) {
-	log.Info("cleaned up syslogNGConfigRef", "name", logging.Status.SyslogNGConfigName)
+	log.Info("cleaning up syslogNGConfig reference from status", "name", logging.Status.SyslogNGConfigName)
 	logging.Status.SyslogNGConfigName = ""
 	err := r.Status().Update(ctx, logging)
 
--- a/controllers/logging/logging_controller.go
+++ b/controllers/logging/logging_controller.go
@@ -30,9 +30,11 @@ import (
 	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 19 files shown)


controllers/logging/logging_controller.go
```diff
@@ -128,6 +130,18 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return reconcile.Result{}, errors.WrapIfWithDetails(err, "failed to get logging resources", "logging", logging)
 	}
 
+	if ExternalFluentdRef := slices.Index(logging.Finalizers, logging.Status.FluentdConfigName); ExternalFluentdRef != -1 {
```
pepov, line 133: I'm missing the deletiontimestamp check from here

controllers/logging/logging_controller.go
```diff
@@ -128,6 +130,18 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		finalizer := logging.Finalizers[ExternalFluentdRef]
+		fluentdConfigName := types.NamespacedName{Namespace: logging.Spec.ControlNamespace, Name: finalizer}
+
+		fluentdConfigRefCleaned := r.cleanupFluentdConfigReference(ctx, log, &logging, fluentdConfigName)
+
+		if !fluentdConfigRefCleaned && logging.DeletionTimestamp != nil {
```
pepov, line 139: why do we need the deletiontimestamp check here?

pkg/resources/model/reconciler.go
```diff
@@ -235,6 +235,7 @@ func NewValidationReconciler(
 			}
 			logger.Info("found detached fluentd aggregator, making association", "name", resources.Fluentd.Configuration.Name)
 			resources.Logging.Status.FluentdConfigName = resources.Fluentd.Configuration.Name
+			resources.Logging.Finalizers = append(resources.Logging.Finalizers, resources.Fluentd.Configuration.Name)
```
pepov, line 238: I'm not sure this is the right place to do this. finalizer operations should be done at a single place

controllers/logging/logging_controller.go
pepov, line 133: also we should put up the finalizer somewhere here I beleive to keep it in one place

OverOrion:
I have reworked the PR according to pepov's notes and using kubebuilder's docs: https://book.kubebuilder.io/reference/using-finalizers.

I also added a secondary fix which occured when logging switched from `fluentdConfig` to inline spec.

controllers/logging/logging_controller.go
```diff
@@ -274,10 +285,53 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			controllerutil.AddFinalizer(logging, fluentdConfigFinalizer)
+			if err := r.Update(ctx, logging); err != nil {
+				return ctrl.Result{}, nil
+			}
+		}
+	} else {
+		// Marked for deletion, check for fluentdConfig
+		fluentdConfigName := types.NamespacedName{Namespace: logging.Spec.ControlNamespace, Name: logging.Status.FluentdConfigName}
+		if controllerutil.ContainsFinalizer(logging, fluentdConfigFinalizer) && r.isExistingFluentdConfig(ctx, fluentdConfigName) {
+			return reconcile.Result{}, errors.NewWithDetails("failed to delete logging resources, delete fluentdConfig first", "fluentdConfig", logging.Status.FluentdConfigName)
```
pepov, line 309: I'm thinking about this hard because the user won't have any feedback about this. How about one of the following?
- we could actually remove the fluentdconfig resource here
- we could update the logging status so that the user see what's happening

or maybe it would be enough to use a logging ownerref on the fluentdconfig?

controllers/logging/logging_controller.go
```diff
@@ -274,10 +285,53 @@ func (r *LoggingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+		controllerutil.RemoveFinalizer(logging, fluentdConfigFinalizer)
+		if err := r.Update(ctx, logging); err != nil {
```
pepov, line 312: what happens to the error here?

kristofgyuracz (approved):

controllers/logging/logging_controller.go
OverOrion: We can not update the status once an object is marked for deletion: https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/#how-finalizers-work (see the note section)

Removing the fluentdconfig might be unexpected (admin deletes logging, tenant loses it's fluentdconfig object)

controllers/logging/logging_controller.go
```diff
@@ -332,6 +376,14 @@ func (r *LoggingReconciler) isExistingFluentdConfig(ctx context.Context, fluentd
 	return true
 }
 
+func (r *LoggingReconciler) isExistingSyslogNGConfig(ctx context.Context, syslogNGConfigref types.NamespacedName) bool {
+	var fluentdConfig loggingv1beta1.SyslogNGConfig
+	if err := r.Client.Get(ctx, syslogNGConfigref, &fluentdConfig); err != nil && apierrors.IsNotFound(err) {
```
orymate, line 381:
```suggestion
	var config loggingv1beta1.SyslogNGConfig
	if err := r.Client.Get(ctx, syslogNGConfigref, &config); err != nil && apierrors.IsNotFound(err) {
```

orymate (approved):

tarokkk (approved):

controllers/logging/logging_controller.go
```diff
@@ -302,15 +309,21 @@ func (r *LoggingReconciler) checkFluentdConfigFinalizer(ctx context.Context, log
...
 		if err := r.Update(ctx, logging); err != nil {
-			return ctrl.Result{}, nil
+			return r.cleanupFluentdConfigReference(ctx, logging, log)
 		}
 	}
 	return ctrl.Result{}, nil
 }
 
+func (r *LoggingReconciler) cleanupFluentdConfigReference(ctx context.Context, logging *loggingv1beta1.Logging, log logr.Logger) (ctrl.Result, error) {
+	log.Info("cleaned up fluentdConfigRef", "name", logging.Status.FluentdConfigName)
```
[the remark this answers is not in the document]
pepov: reworded

pepov (approved):

PR #1682, opened 2024-03-01
Isolate input sources to avoid backpressure affecting other tenants

pepov:
The consequences of a faulty output can effect other tenants if backpressure kicks in. To avoid this problem we separate inputs from each other by creating a dedicated tail input for each tenant.


tarokkk (approved):
Great insights! Thanks

OverOrion (approved):

PR #1685, opened 2024-03-01
Add support for VMware Log Intelligence

zrobisho:
- Add support for VMware Log Intelligence fluentd plugin https://github.com/vmware/fluent-plugin-vmware-log-intelligence


pkg/sdk/logging/model/output/vmware_log_intelligence.go
```diff
@@ -0,0 +1,151 @@
...
+	verify_ssl: true
+	http_compress: false
+	headers:
+	  Content-Type: "application/json"
+	  Authorization: "Bearer 12345"
+	  structure: simple
+	buffer:
+	  chunk_limit_records: 300
+	  flush_interval: 3s
+	  retry_max_times: 3
```
pepov, line 43: indentation is a bit off here
```suggestion
    endpoint_url: https://data.upgrade.symphony-dev.com/le-mans/v1/streams/ingestion-pipeline-stream
    verify_ssl: true
    http_compress: false
    headers:
      Content-Type: "application/json"
      Authorization: "Bearer 12345"
      structure: simple
    buffer:
      chunk_limit_records: 300
      flush_interval: 3s
      retry_max_times: 3
```

pkg/sdk/logging/model/output/vmware_log_intelligence.go
```diff
@@ -0,0 +1,151 @@
...
+// +status:"GA"
+type _metaVMWareLogIntelligence interface{} //nolint:deadcode,unused
+
+// +kubebuilder:object:generate=true
+// +docName:"VMwareLogIntelligence"
+type VMWareLogIntelligenceOutput struct {
+	// Log Intelligence endpoint to send logs to https://github.com/vmware/fluent-plugin-vmware-log-intelligence?tab=readme-ov-file#label-endpoint_url
+	EndpointURL string `json:"endpoint_url"`
+	// Verify SSL (default: true) https://github.com/vmware/fluent-plugin-vmware-log-intelligence?tab=readme-ov-file#label-verify_ssl
+	VerifySSL bool `json:"verify_ssl" plugin:"default:true"`
```
pepov, line 61: If we add a default: true, then you won't be able to disable it. Please use a pointer to enable users disabling it.
```suggestion
	VerifySSL *bool `json:"verify_ssl" plugin:"default:true"`
```

pkg/sdk/logging/model/output/vmware_log_intelligence.go
```diff
@@ -0,0 +1,151 @@
...
+	// Compress http request https://github.com/vmware/fluent-plugin-vmware-log-intelligence?tab=readme-ov-file#label-http_compress
+	HTTPCompress *bool `json:"http_compress,omitempty"`
```
pepov, line 63: Nit: this doesn't have to be a pointer, since it's false by default (although it doesn't hurt either)

pepov (changes requested):
Just a few minor things, thanks for the PR!

pkg/sdk/logging/model/output/vmware_log_intelligence.go
zrobisho: Thanks, my editor was hiding some tabs it seems

pkg/sdk/logging/model/output/vmware_log_intelligence.go
zrobisho: The reason I used  a pointer was so that when the user explicitly sets `false` in the `Output` the field is still rendered correctly in the fluentd config and not removed as I figured this was more consistent for the user creating the `Output`.

zrobisho:
@pepov I've addressed the comments, please lmk if there is anything else you'd like me to look at. Thanks for the review!

pepov (approved):

OverOrion:
Hey @zrobisho,

Thank you for your PR, could you rebase it against the latest master? I think running `make generate` should do the trick.

zrobisho:
@OverOrion Rebased

OverOrion (approved):

PR #1699, opened 2024-03-14
chore: remove obsolete deprecation warning from logging status

### LINKED ISSUE #1666: Obsolete deprecation warning
**Describe the bug**:
The following deprecation warning is no longer accurate:

> Deprecated behaviour! Other logging resources exist with the same loggingRef: example. This is going to be an error with the next major release.

**Expected behaviour**:
This deprecation should only be a notice, as there are certain use cases, where this behaviour is otherwise appropriate. 

Create two logging resources without specifying a loggingRef or by specifying the same loggingref and check their status.

pepov:
make the conflict warning smarter and more informative

fixes #1666

### DIFF
--- a/pkg/resources/model/reconciler.go
+++ b/pkg/resources/model/reconciler.go
@@ -34,6 +34,8 @@ import (
 	"github.com/kube-logging/logging-operator/pkg/mirror"
 )
 
+const LoggingRefConflict = "Other logging resources exist with the same loggingRef"
+
 func NewValidationReconciler(
 	repo client.StatusClient,
 	resources LoggingResources,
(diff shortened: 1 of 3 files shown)


pkg/resources/model/reconciler.go
```diff
@@ -417,3 +417,14 @@ func jsonFieldName(f reflect.StructField) string {
 	}
 	return f.Name
 }
+
+func hasIntersection(a, b []string) bool {
```
OverOrion, line 423: [nit]
If the slices have many element, then maybe a `map[string]struct{}` would be better here, but let's cross that bridge once we get there.

OverOrion (approved):
LGTM! :rocket:

PR #1704, opened 2024-03-20
e2e test for config checks

### LINKED ISSUE #1540: Create e2e test for config checks
Config check failure on fluentd creating a file output with an invalid pattern:
```
spec:
  file:
    append: true
    buffer:
      timekey: 1m
      timekey_use_utc: true
      timekey_wait: 10s
    path: /tmp/logs/asd
```

Output
```
2023-10-16 13:59:22 +0000 [error]: config error file="/fluentd/etc/fluent.conf" error_class=Fluent::ConfigError error="Parameter 'path: /tmp/logs/asd.%Y%m%d%H%M.log' doesn't have tag placeholder"
```

Test scenario to implement:
- Creating a valid configuration succeeds and writes result into the logging resource status.
- Updating to an invalid configuration modifies the status with a failed result.
- Reverting back get passed the configcheck.

TODO
- [ ] define example for syslog-ng as well

nak0f:
Fixes #1540

### DIFF
--- a/e2e/fluentd-aggregator/fluentd_aggregator_test.go
+++ b/e2e/fluentd-aggregator/fluentd_aggregator_test.go
@@ -223,7 +223,7 @@ func TestFluentdAggregator_ConfigChecks(t *testing.T) {
 	releaseNameOverride := "e2e"
 	outputName := "test-output"
 	flowName := "test-flow"
-	common.WithCluster("fluentd-1", t, func(t *testing.T, c common.Cluster) {
+	common.WithCluster("fluentd-2", t, func(t *testing.T, c common.Cluster) {
 		setup.LoggingOperator(t, c, setup.LoggingOperatorOptionFunc(func(options *setup.LoggingOperatorOptions) {
 			options.Namespace = ns
 			options.NameOverride = releaseNameOverride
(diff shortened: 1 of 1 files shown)


e2e/fluentd-aggregator/fluentd_aggregator_test.go
```diff
@@ -217,3 +217,176 @@ func TestFluentdAggregator_MultiWorker(t *testing.T) {
...
+			return logging.Status.ProblemsCount == 0
+		}, 5*time.Minute, 3*time.Second)
+
+		t.Logf("Breaking File Output with an invalid config")
+		patch := client.MergeFrom(output.DeepCopy())
+		output.Spec.FileOutput.Path = "/tmp/zzz"
+		common.RequireNoError(t, c.GetClient().Patch(ctx, &output, patch))
+		require.Eventually(t, func() bool {
+			common.RequireNoError(t, c.GetClient().Get(ctx, utils.ObjectKeyFromObjectMeta(&logging), &logging))
+			if logging.Status.ProblemsCount != 1 {
```
pepov, line 350: @nak0f this is great, but what if there is another, yet unrelated problem reported here instead of the one we are aiming for? We could check if there is a failed pod with the configcheck labels, what do you think?

nak0f: @pepov : Thanks for taking a look into this. Good call for the problem checking part. I've just pushed a fix to address this concern.

pepov (approved):

e2e/fluentd-aggregator/fluentd_aggregator_test.go
pepov: nit: I'm still a bit concerned about checking the error message instead of the expected result (there is a failed configcheck pod) but it's still better than no check at all

OverOrion (approved):

PR #1734, opened 2024-05-09
chore(controller-gen): update controller-gen and add changes after running it

kristofgyuracz:


Makefile
```diff
@@ -140,8 +140,8 @@ manager: codegen fmt vet ## Build manager binary
 
 .PHONY: manifests
 manifests: ${CONTROLLER_GEN} ## Generate manifests e.g. CRD, RBAC etc.
-	cd pkg/sdk && $(CONTROLLER_GEN) $(CRD_OPTIONS) webhook paths="./..." output:crd:artifacts:config=../../config/crd/bases output:webhook:artifacts:config=../../config/webhook
```
pepov, line 143: I don't think we can remove CRD_OPTIONS, as for example `maxDescLen=0` is limiting our crds in size. I attribute at least some of the test failures to this.

kristofgyuracz: Added maxdesclen, others are deprecated.

pepov (approved):

PR #1756, opened 2024-06-07
syslogng: render go templates in metrics-probe label values

orymate:
This PR adds flexibility in what label keys should be used automatically for output metrics, also adds the ability to use go templates that receives the destination information (name, namespace, scope:local/global and the logging name) to be used as needed. Example:
```
kind: SyslogNGClusterFlow
metadata:
  name: all
spec:
  match: {}
  outputMetrics:
    - key: custom_output
      labels:
        flow: all
        # define your own label for output name
        my-key-for-the-output: "{{ .Destination.Name }}"
        # do not add the output_name label to the metric
        output_name: ""
  globalOutputRefs:
    - http
```

### DIFF
--- a/pkg/sdk/logging/model/syslogng/config/log.go
+++ b/pkg/sdk/logging/model/syslogng/config/log.go
@@ -34,11 +34,11 @@ const (
 )
 
 type destination struct {
-	logging          string
+	Logging          string
 	renderedDestName string
 	Namespace        string
 	Name             string
-	scope            refScope
+	Scope            refScope
 	metricsProbes    []filter.MetricsProbe
 }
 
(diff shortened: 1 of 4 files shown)


pepov (approved):

pkg/sdk/logging/model/syslogng/config/log.go
```diff
@@ -80,3 +99,20 @@ func destinationLogPath(dest destination) render.Renderer {
 		parenDefStmt("destination", render.Literal(dest.renderedDestName))),
 	)
 }
+
+func renderLabelGoTemplates(labels filter.ArrowMap, values any) filter.ArrowMap {
+	for k, v := range labels {
```
orymate, line 104:
```suggestion
	for k, v := range maps.Clone(m.Labels) {
```

this could come here, so we explicitly return a changed copy

OverOrion (approved):

PR #1759, opened 2024-06-13
fix: added missing log section inside system config 

pbdger:
The new feature logFormat text | json did not work as expected.

This is caused by a misconfiguration inside the system config section. This pull request fix the issue.

For further information see https://docs.fluentd.org/deployment/system-config#less-than-log-greater-than-section


pbdger:
Will be done in a new pull-request

PR #1760, opened 2024-06-14
fix: added missing log section inside system config

pbdger:
see  https://docs.fluentd.org/deployment/system-config#less-than-log-greater-than-section

### DIFF
--- a/pkg/resources/fluentd/config.go
+++ b/pkg/resources/fluentd/config.go
@@ -33,8 +33,10 @@ var fluentdInputTemplate = `
 # Enable RPC endpoint (this allows to trigger config reload without restart)
 <system>
   rpc_endpoint 127.0.0.1:24444
-  {{- if .LogFormat }}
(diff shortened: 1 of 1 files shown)


OverOrion (approved):
Thank you for your contribution @pbdger!

pkg/resources/fluentd/config.go
```diff
@@ -33,8 +33,10 @@ var fluentdInputTemplate = `
 # Enable RPC endpoint (this allows to trigger config reload without restart)
 <system>
   rpc_endpoint 127.0.0.1:24444
-  {{- if .LogFormat }}
-  format {{ .LogFormat }}
+  {{- if .LogFormat }}  
+  <log>
+    format {{ .LogFormat }}
+  </log>format {{ .LogFormat }}
```
aslafy-z, line 39:
```suggestion
  </log>
```
I feel like this should be removed.

pepov (approved):

Issue #1749, opened 2024-05-27
Make severity configurable in PrometheusRules


sebastiangaiser:
**Is your feature request related to a problem? Please describe.**
Currently the severity is hardcoded in the PrometheusRules like [here](https://github.com/kube-logging/logging-operator/blob/a71e24e740c893c92c364064647e5bf9070f2aa8/pkg/resources/fluentd/prometheusrules.go#L47)

**Describe the solution you'd like**
It would be great to have them configurable via the Logging CR e.g.:

```
spec:
  fluentd:
    metrics:
      prometheusRules: true
      severity: my-custom-severity
  fluentbit:
    metrics:
      prometheusRules: true
      severity: my-custom-severity
  syslogNG:
    metrics:
      prometheusRules: true
      severity: my-custom-severity
```

**Describe alternatives you've considered**
My current workaround is to disable the PrometheusRules and deploy them in the way I want

**Additional context**
Logging Operator docs: https://kube-logging.dev/docs/operation/alerting/

sebastiangaiser:
@pepov sorry to ping you but could you please mind have a look :D

pepov:
Good idea, but I don't have any free cycles currently. We are looking for contributors to take on "good first issues" like this!

aditya7302:
I would like to work on this issue. Can you assign it to me ?

pepov:
@aditya7302 sure, thanks!

aditya7302:
@pepov is this [file](https://github.com/kube-logging/logging-operator/blob/a71e24e740c893c92c364064647e5bf9070f2aa8/config/crd/bases/logging.banzaicloud.io_loggings.yaml) is where severity metric should be added to make it configurable ?

sebastiangaiser:
@pepov @aditya7302 maybe my comment now is a bit controversial to remove the PrometheusRules completely or making the whole [Rules block configurable](https://github.com/kube-logging/logging-operator/blob/3c20c69c03f588a629b89e92f1b871ebb8e77038/pkg/resources/fluentbit/buffervolumeprometheusrules.go#L40)

pepov:
I would add a list based override option that would allow setting overrides for the rules based on the list index. Also we could define a type that would allow overriding other fields as well for the rule.

For example the below metrics config would keep the first rule unmodified but would change a few properties of the second rule.
```
    metrics:
      prometheusRules: true
      prometheusRuleOverrides: 
        - {}
        - for: 1m
          labels:
[...]

PR #1768, opened 2024-07-01
feat(charts,logging-operator): add possibility to deploy extra manife…

sebastiangaiser:
…sts to Helm chart


sebastiangaiser:
@pepov maybe you can include that in the upcomming release tomorrow :D

pepov:
@sebastiangaiser what is the usecase for this? what is the reason against creating an umbrella chart and putting extra resources there?

sebastiangaiser:
My use case are the PrometheusRules from [here](https://github.com/kube-logging/logging-operator/issues/1749)

pepov:
I meant that you can create a high level chart, add you prometheus resources and add the logging operator as a dependency into that. I haven't really seen this extra resources as a practice this is why I ask.

The term comes from here: https://helm.sh/docs/howto/charts_tips_and_tricks/#complex-charts-with-many-dependencies

pepov:
I've found a few usage for this in the most popular charts, ok for me

pepov (approved):

OverOrion (approved):

PR #1769, opened 2024-07-02
chore: remove trace log setting from examples to avoid suggesting a bad practice

pepov:


OverOrion (approved):

jszigetvari (approved):
Great, thank you!

PR #1773, opened 2024-07-04
fix: direct map of fluentbit securitycontext options

aslafy-z:
- Send whole fluentbit SecurityContext to fluentbit, configReloader and bufferVolumeMetrics containers


pepov:
@aslafy-z perhaps set the whole `spec.Security.SecurityContext` directly into `c.SecurityContext`?

aslafy-z:
I updated the PR, should I map it to buffervolumemetrics container too?

pepov:
@aslafy-z I beleive yes, that would be symmetric to what we do with fluentd's bufferVolumeMetrics container as well

pepov (approved):

OverOrion (approved):

PR #1828, opened 2024-10-14
OpenSearch remove_keys parameter support

kampka:
Adds support for [remove_keys](https://github.com/fluent/fluent-plugin-opensearch?tab=readme-ov-file#remove_keys) to the opensearch plugin.


sebastiangaiser:
@pepov would be great if we could get this for https://github.com/kube-logging/logging-operator/milestone/31

pepov:
@kampka nothing to be done, thanks for the contribution!

sebastiangaiser:
Thank you @pepov 🚀

PR #1831, opened 2024-10-20
fix(syslogng_servicemonitor): add missing fields from syslogng-spec

csatib02:


OverOrion (approved):
LGTM, thanks for spotting it!

PR #1833, opened 2024-10-23
fix(charts,logging-operator): add empty workloadMetaOverrides field i…

sebastiangaiser:
…n hosttailer if not specified

The '.spec.workloadMetaOverrides' field is required for all hosttailer resources. Unfortunately if not set the deployment via the chart fails with 'is invalid: spec.workloadMetaOverrides: Required value'. In order to mitigate this, an empty object is injected if not specified.


sebastiangaiser:
@csatib02 please don't merge this. Even passing an empty object isn't allowed, too. I think this might need to be fixed in the operator and relaxing the crd.

sebastiangaiser:
@csatib02 I created a bug instead, sorry for the turn around.

PR #1836, opened 2024-10-30
expose fluentd metrics also on IPv6

### LINKED ISSUE #1835: fluentd metrics cannot be scraped on IPv6 only clusters
**Describe the bug**:

fluentd by default only opens the metrics port on `0.0.0.0:24231`, ie. not exposing it for IPv6.

Upstream issue: https://github.com/fluent/fluentd/issues/3001 (it seems complicated though, because a library needs to be updated/switched)

A long-open PR for the fluentd helm repository (https://github.com/fluent/helm-charts/pull/447) proposes just adding a second Prometheus source. Seems a valid aproach to me, given fluentd is running in a container and nobody will have blocked the IPv6 port as feared by the upstream issue, also it was originally denied (https://github.com/fluent/fluent-plugin-prometheus/issues/151).

logging-operator currently [does not expose configuration for the required monitoring settings](https://github.com/kube-logging/logging-operator/blob/master/pkg/resources/fluentd/config.go#L59-L65):

```
<source>
    @type prometheus
    port {{ .Monitor.Port }}
{{- if .Monitor.Path }}
    metrics_path {{ .Monitor.Path }}
{{- end }}
</source>
```

Possible solutions I can think of:

- generally open IPv6 like the helm chart PR above does https://github.com/kube-logging/logging-operator/pull/1836
- add a bind configuration option to the [metrics type](https://github.com/kube-logging/logging-operator/blob/master/pkg/sdk/logging/api/v1beta1/common_types.go#L66-L77); kind of straight forward and would allow to "choose" between IPv4/IPv6, but needs handling the option for the other implementations (syslog tailer, fluentbit) -- I consider this feasible and we would provide a PR, but needs discussion first
- Add a "generic" configuration injection option (unless there is one we missed), such that one can cleanly inject a matching configuration block. Not really a nice solution for this issue, but probably generally good to have for special use cases and workarounds.
- ...?

**Expected behaviour**:

Metrics are either available on IPv4 and IPv6 by default (preferred, as "just works"), or the behavior can be toggled, such that IPv6 only clusters can scrape fluentd metrics.

Steps to reproduce the bug should be clear and easily reproducible to help people
gain an understanding of the problem.

**Additional context**:

**Environment details**:
- Kubernetes version (e.g. v1.15.2):
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc):
- logging-operator version (e.g. 2.1.1): 4.10.0/from master

/kind bug

JensErat:
Currently, IPv6 only clusters cannot scrape metrics from fluentd, as fluentd's `bind` default is `0.0.0.0`. Any IPv6 only cluster will fail scrapes.

fluentd currently does not support binding both IPv6 and IPv4 in a single source. Adding a second IPv6 source makes fluentd listen on both IPv4 and IPv6, allowing scrapes universally.

Fixes #1835


csatib02 (approved):
LGTM!

pepov (approved):

PR #1840, opened 2024-11-04
Possibility to add strategy and topologySpreadConstraints to the operator

karloramone:
If you have two datacenters and you would like to deploy two replicas of the operator in each, you can't do that cause there is no possibility to modify topologySpreadConstraints of the operator deployment. The only way to assure two pods running in each datacenters is to run three replicas with the podAntiAffinity.

If you have a three node cluster you are not able to run three replicas with podAntiAffinity on the hostname topology key.
The problem is in the default strategy is configured to add the fourth replica before removing the old one. Because of the mentioned configuration above this is not possible and pod stays in pending state.

This fixes those two examples.
There is also one typo on the README.md that I noticed and fixed.


csatib02 (approved):
LGTM!

Please run `make generate` so the CI can pass.

karloramone:
Thanks @csatib02, I did it.
However, two jobs failed.
vulndb would probably fixed with the re-run and for the E2E tests I'm not sure.
Can you re-run the  pipeline?

PR #1844, opened 2024-11-07
fix fluentbit ipv6 listen address

eumel8:
if logging-operator is deployed in IPv6 environment, fluentbit is not able to listen on http socket due the hard coded `HTTP_Listen` address. Make this dynamically set by `POD_IP` with an envVar definition

### DIFF
--- a/pkg/resources/fluentbit/config.go
+++ b/pkg/resources/fluentbit/config.go
@@ -35,7 +35,7 @@ var fluentBitConfigTemplate = `
     Coro_Stack_Size    {{ .CoroStackSize }}
     {{- if .Monitor.Enabled }}
     HTTP_Server  On
-    HTTP_Listen  ${POD_IP}
+    HTTP_Listen  [::]
     HTTP_Port    {{ .Monitor.Port }}
     {{- end }}
     {{- range $key, $value := .BufferStorage }}
--- a/pkg/resources/fluentbit/config.go
+++ b/pkg/resources/fluentbit/config.go
@@ -35,7 +35,7 @@ var fluentBitConfigTemplate = `
     Coro_Stack_Size    {{ .CoroStackSize }}
     {{- if .Monitor.Enabled }}
     HTTP_Server  On
-    HTTP_Listen  0.0.0.0
+    HTTP_Listen  ${POD_IP}
     HTTP_Port    {{ .Monitor.Port }}
     {{- end }}
     {{- range $key, $value := .BufferStorage }}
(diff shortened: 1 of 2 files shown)


pkg/resources/fluentbit/config.go
```diff
@@ -35,7 +35,7 @@ var fluentBitConfigTemplate = `
     Coro_Stack_Size    {{ .CoroStackSize }}
     {{- if .Monitor.Enabled }}
     HTTP_Server  On
-    HTTP_Listen  0.0.0.0
```
eumel8, line 38: another way: 

```
HTTP_Listen  [::]
```

works on IPv6 and doesn't need the Statefulset patch

pepov:
I tried to rerun the e2e test multiple times, but it fails, haven't been able to dig deeper why. @eumel8 do you have an idea what can be the problem?

pkg/resources/fluentbit/config.go
eumel8: I mean the addional env var on $POD_IP in the Statefulset manifest. But this is not required anymore. Tested on Ipv6 and Ipv4 cluster the definition of `[::]` works.
The end2end test failed of the change in Statefulset manifest, but this should fixed now.

csatib02:
I believe this PR is somewhat similiar to the one we merged not long ago: https://github.com/kube-logging/logging-operator/pull/1836

eumel8:
@csatib02 Yes, after implementation we realized in an IPv6-only-cluster that fluentbit would listen only on IPv4 socket.

pepov (approved):

withlin:
hi @eumel8  @pepov  shall we support set a enbale for ipv6 or ipv4? because somebody  do not support ipv6 yet  on machine for security.

csatib02:
Hi @withlin,

Thank you for your feedback regarding IPv6 and IPv4 support for Fluentbit and Fluentd!

Adding the ability to configure this behavior at the operator level sounds like a reasonable enhancement. Contributions in this area are very welcome, as they would help address this use-case.

If you’re interested, feel free to open a pull request—we’d be happy to collaborate on it!

eumel8:
Hi @withlin ,

yes, I had this in mind and [my previous approach](https://github.com/eumel8/logging-operator/compare/master...eumel8:logging-operator:fix/fluentbit-ipv6-2) was to use POD_IP instead of catch all listen address. Just tested on IPv4 cluster I've got this error

```
  Warning  FailedCreate      84s (x15 over 2m44s)  daemonset-controller  Error creating: pods "infra-fluentbit-" is forbidden: Get "https://[::1]:6443/api/v1/namespaces/infra/limitranges": dial tcp [::1]:6443: connect: cannot assign requested address
```

so, correct way would be to open issue and we can work on bug fix

PR #1845, opened 2024-11-08
fix: configcheck cleaner label selector

pepov:
ConfigCheck cleaner was not aware of the logging resource and cleaned all configcheck secrets and pods that belonged to other logging resources. Adding the managed-by label to the cleaner filter solves this problem.


csatib02 (approved):
LGTM! :shipit:

PR #1849, opened 2024-11-11
chore(pkg): update AxoSyslog image to 4.9.0

MrAnno:


pkg/sdk/logging/api/v1beta1/syslogng_types.go
```diff
@@ -34,7 +34,7 @@ type _metaSyslogNGSpec interface{} //nolint:deadcode,unused
 
 const (
 	defaultSyslogngImageRepository    = "ghcr.io/axoflow/axosyslog"
-	defaultSyslogngImageTag           = "4.8.1-1"
+	defaultSyslogngImageTag           = "4.9.0-1"
```
pepov, line 37: Can we use the moving that, so that we can leverage occassional base image vuln fixes? We have the same process for fluentd as well.
```suggestion
	defaultSyslogngImageTag           = "4.9.0"
```

MrAnno: Yep, done :)

PR #1857, opened 2024-11-25
Telemetry Controller compatibility example

pepov:
Add an example on how the telemetry controller resources compare to a similar legacy logging operator setup regarding log collection and routing.


OverOrion (approved):
Nice, thanks!

Issue #1798, opened 2024-08-22
Reintroduce Openshift SCC to the logging-operator chart


pepov:
> Bugs should be filed for issues encountered whilst operating logging-operator.
> You should first attempt to resolve your issues through the community support
> channels, e.g. Slack, in order to rule out individual configuration errors. [#logging-operator](https://slack.banzaicloud.io/)
> Please provide as much detail as possible.

A user noticed that the OpenShift SCC resources disappeared and would like to still have it: https://github.com/kube-logging/logging-operator/pull/1167

pepov:
I just realized, that this is something to be added to the actual logging operator code, it doesn't have a good place in the chart as it belongs to the aggregator instead of the logging operator deployment itself. I have to defer this to the next release.

PR #1167, opened 2023-01-22
Add support for OpenShift SCC

aelbarkani:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1166
| License         | Apache 2.0

### What's in this PR?
Add the possibility to deploy an SCC in OpenShift in order to allow Fluentbit to run in privileged mode.

### Why?
Solves #1166, to avoid deploying SCC separately.

### Checklist

- [x] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [x] User guide and development docs updated (if needed)
- [x] Related Helm chart(s) updated (if needed)

PR #1862, opened 2024-11-27
fix: re-introduce scc

csatib02:
Fixes: #1798


OverOrion (approved):
Just for future reference this PR re-introduces the behavior in https://github.com/kube-logging/logging-operator/pull/1167

Issue #1327, opened 2023-05-07
Overly long generated names


pepov:
**Describe the bug**:
Resource names generated by the operator can be too long and refused by kubernetes. 

**Expected behaviour**:
The operator should not generate names longer then the allowed resource name limit, but rather should have a way to generate uniq names for those cases as well

**Steps to reproduce the bug**:
Use the `logging-operator` chart and the generated pvc name will be too long.

See https://emergingtechcommunity.slack.com/archives/CGMPPBZBK/p1683467594142789

PR #1863, opened 2024-11-27
fix: long generated names

csatib02:
### Overview

According to [k8s's docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) resource names 

1. contain at most 63 characters
2. contain only lowercase alphanumeric characters or '-'
3. start with an alphanumeric character
4. end with an alphanumeric character
5. be unique across all [API versions](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning) of the same resource. API resources are distinguished by their API group, resource type, namespace (for namespaced resources), and name. In other words, API version is irrelevant in this context.

#### Name-checking explanation

First of all we make use of the built-in validation function provided by k8s, if that fails, we do the following:

1. Hashing for Uniqueness (if needed):

- When the combined name exceeds 253 characters, the function adds an 8-character hash derived from the full name.
- This ensures that truncated names remain unique and avoids conflicts.

2. Truncation (if needed):

- The baseName is truncated to leave space for the hash and hyphen.
- The truncation ensures the total length of the generated name stays within 253 characters.

3. Edge Case Handling:

- If the name length is already within the limit, it is returned as-is without modification.

NOTE: Also fixed a typo in a file's name.

Should fix: #1327


pepov:
After discussing this we concluded to use a simple length check with 40 characters and see if that is enough to avoid failures due to not just resource names but label value constraints as well.

csatib02:
@pepov I had to set the `max-length` to 50, because setting it to 40 produced crazy amount of errors around tests.

Issue #1859, opened 2024-11-27
Remove resource finalizers before terminating operator


mriley-mbgp:
**Is your feature request related to a problem? Please describe.**
For cluster consistency, we have developed a Helm chart to install the logging-operator as a dependency and deploy our configuration files tweaked as necessary. We have used the new recommended pattern of separate custom resources for the configuration by utilising `SyslogNGConfig` in our deployment, which adds a finalizer onto the matching `Logging` resource to prevent deletion of this resource.

```
Name:         dev-cluster-logging-agent
Namespace:
Labels:       app.kubernetes.io/instance=dev-cluster-logging
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=dev-cluster-logging
              app.kubernetes.io/version=0.2.4
              helm.sh/chart=mbgp-cluster-logging-0.1.0
Annotations:  fluentbit.io/exclude: true
              meta.helm.sh/release-name: dev-cluster-logging
              meta.helm.sh/release-namespace: logging
API Version:  logging.banzaicloud.io/v1beta1
Kind:         Logging
Metadata:
  Creation Timestamp:  2024-11-22T14:03:57Z
  Finalizers:
    syslogngconfig.logging.banzaicloud.io/finalizer
  Generation:        2
  Resource Version:  2339510619
  UID:               a59cea21-b4f8-4cb0-bed8-6bc0fea99392
Spec:
  Cluster Domain:  cluster.local.
  Config Check:
    Timeout Seconds:  10
  Control Namespace:  logging
Status:
  Config Check Results:
    6d39d335:             true
  Syslog NG Config Name:  dev-cluster-logging-agent
Events:                   <none>
```

When manually removing the `SyslogNGConfig` only, the finalizer is successfully removed by the logging operator off the `Logging` resource. When performing a Helm uninstall, the logging-operator terminates without removing finalizers, so the `Logging` resource is left dangling which can persist old configuration when deploying over the top of an old deployment. 

**Describe the solution you'd like**
It's hard to know exactly how one should proceed solution wise, but an idea might be to check for existing finalizers before termination of the logging operator and remove them. As the logging operator is the one that manages the allocation and checks around finalizers, if this resource doesn't exist, it will only cause dangling resources. This method could lead to an "at risk" period if the logging operator were to crash or be rescheduled, before returning them on next start up.

**Describe alternatives you've considered**
[...]

PR #1872, opened 2024-11-29
feat: add pre-delete-hook

csatib02:
Fixes: #1859


charts/logging-operator/templates/pre-delete-hook.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: batch/v1
```
aslafy-z, line 1: I think this job should be opt in:
- uninstall does not mean delete every CRs
- argocd does not support pre-delete hook

PR #1873, opened 2024-11-30
feat: add option to cleanup stuck finalizers

csatib02:
## Overview

This PR aims to address the issue raised in #1859.

### Problem

The issue arises because Helm does not guarantee a precise order in which it deletes resources. Although this order can be influenced via hooks and weights, there are limitations. For example, hooks may not work as expected in ArgoCD environments.

### Proposed Solution

- Added a new flag to the operator service: `finalizer-cleanup`. This flag should be used in conjunction with `.Values.rbac.retainOnDelete`.
- When both options are set and `helm uninstall` is executed:
  1. Helm will retain the operator's service account as well as its cluster role and cluster role binding. This is necessary because Helm uninstalls resources in an unpredictable order.
  2. The operator will attempt to free up its managed logging resources by removing finalizers, which would otherwise remain stuck.

### How to Use This Feature

Since this problem was observed when the operator was uninstalled via `helm uninstall`, the feature has been implemented to be enabled by modifying the chart:

```sh
helm install logging-operator ./charts/logging-operator/ --set rbac.retainOnDelete=true --set extraArgs='{"-enable-leader-election=true","-finalizer-cleanup=true"}'
```

### DIFF
--- a/charts/logging-operator/values.yaml
+++ b/charts/logging-operator/values.yaml
@@ -60,6 +60,8 @@ http:
 rbac:
   # -- Create rbac service account and roles.
   enabled: true
+  # -- Keep the operators RBAC resources after the operator is deleted to allow removing pending finalizers.
+  retainOnDelete: false
 
   # specify service account manually
   # serviceAccountName: custom
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,8 @@ list: ## List all make targets
 manager: codegen fmt vet ## Build manager binary
 	go build -o bin/manager main.go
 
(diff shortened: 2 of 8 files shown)


charts/logging-operator/values.yaml
```diff
@@ -60,6 +60,8 @@ http:
 rbac:
   # -- Create rbac service account and roles.
   enabled: true
+  # -- Keep the operators RBAC resources after the operator is deleted.
```
pepov, line 63:
```suggestion
  # -- Keep the operators RBAC resources after the operator is deleted to allow removing pending finalizers.
```

pepov (approved):

PR #1876, opened 2024-12-03
feat: add TC migration option

csatib02:
## Overview

- The core addition of this PR is to enable log collection by utilizing the [Telemetry-Controller](https://github.com/kube-logging/telemetry-controller) resources. This is done via the `logging` resource.
- There is a new field on `logging` that will determine how TC resources will be created:

```yaml
kind: Logging
metadata:
  name: tc-config
spec:
  routeConfig:
    enableTelemetryControllerRoute: true # Determines whether to create TC resources for log-collection and routing purposes.
    disableLoggingRoute: false # Determines whether to use Logging-routes for routing purposes and Fluentbit-agents for log-collection.
    tenantLabels: # Will be placed on TC's tenant resources, must be matched with the same field on the Collector resource. (Deployed by the cluster administrator.)
      tenant: logging
```
- To use this one most set the `enable-telemetry-controller-route` flag to `true` Eg.:
```sh
helm install logging-operator ./charts/logging-operator/ --set extraArgs='{"-enable-leader-election=true","-enable-telemetry-controller-route"}' --set telemetry-controller.install=true
```

- Adds option to install telemetry-controller as a sub-chart.
- Sets all API occurrences of the `ControlNamespace` field to be immutable.


pkg/sdk/logging/api/v1beta1/logging_types.go
```diff
@@ -114,6 +121,19 @@ type ConfigCheck struct {
 	Labels map[string]string `json:"labels,omitempty"`
 }
 
+type RouteConfig struct {
+	// If DisableLoggingRoute is set to true, the logging route controller
+	// should remove the given tenant from the status of the logging resource.
+	DisableLoggingRoute bool `json:"disableLoggingRoute,omitempty"`
+	// If EnableTelemtryControllerRoute set to true, the opoerator will create
```
OverOrion, line 128:
```suggestion
	// If EnableTelemtryControllerRoute set to true, the operator will create
```

pkg/sdk/logging/api/v1beta1/logging_types.go
```diff
@@ -92,6 +96,9 @@ type LoggingSpec struct {
 	// - downstream parsers can use the `log` field instead of `message` as they did with the docker runtime
 	// - the `concat` and `parser` filters are automatically set back to use the `log` field
 	EnableDockerParserCompatibilityForCRI bool `json:"enableDockerParserCompatibilityForCRI,omitempty"`
+	// RouteConfig determines whether to use loggingRoutes or to create resources based on the logging resource
+	// that can be managed by the Telmetry Controller.
```
OverOrion, line 100:
```suggestion
	// that can be managed by the Telemetry Controller.
```

OverOrion (commented):
Besides 2 typos this looks good to me, nice! :rocket:

pkg/sdk/logging/api/v1beta1/logging_types.go
```diff
@@ -114,6 +121,19 @@ type ConfigCheck struct {
...
+	// If DisableLoggingRoute is set to true, the logging route controller
+	// should remove the given tenant from the status of the logging resource.
+	DisableLoggingRoute bool `json:"disableLoggingRoute,omitempty"`
+	// If EnableTelemtryControllerRoute set to true, the operator will create
+	// the corresponding Tenant, Subscription, Output based on the logging resource.
+	EnableTelemetryControllerRoute bool `json:"enableTelemetryControllerRoute,omitempty"`
+	// TenantLabels is a map of labels that will be added to the tenant object
+	// so it can be matched with TelemetryController's TenantSelector
+	// ref: https://github.com/kube-logging/telemetry-controller/blob/main/api/telemetry/v1alpha1/collector_types.go
+	TenantLabels *metav1.LabelSelector `json:"tenantLabels,omitempty"`
```
csatib02, line 134: We might want to change this to a simple `map[string]string`
@OverOrion @pepov WDYT?

OverOrion (approved):

Issue #1909, opened 2025-01-13
Support multiple HostTailers in the chart


pepov:
The helm chart currently only supports defining a single host tailer, but there are situations where more than one need to be defined, for example when different host tailers need to be deployed to different set of nodes using nodeselectors.

Example:
```
hostTailer:
  enabled: true
  - name: kubeauditane                                 # <--------- currently this is not a list  
    workloadOverrides:
      nodeSelector:
        node-role.kubernetes.io/control-plane: "true"
      tolerations:
        - key: node-role.kubernetes.io/control-plane
          operator: Exists
          effect: NoSchedule
    fileTailers:
      - name: kube-audit
        path: /var/lib/rancher/rke2/server/logs/*.log
  
  - name: workersnodesonly
    workloadOverrides:
      nodeSelector:
        node-role.kubernetes.io/worker: "true"
    fileTailers:
      - name: kube-audit
        path: /var/lib/rancher/rke2/agent/logs/*.log
```

PR #1918, opened 2025-01-20
feat(chart): enable defining multiple hosttailers

csatib02:
Fixes: https://github.com/kube-logging/logging-operator/issues/1909

### Deprecation notice

`logging.hostTailer` is superseded by `logging.hostTailers` and will be removed in a future release

### Example

```yaml
logging:
  enabled: true
  hostTailers:
    enabled: true
    instances:
      - name: kubeauditane
        enabled: true
        workloadOverrides:
          nodeSelector:
            node-role.kubernetes.io/control-plane: "true"
          tolerations:
            - key: node-role.kubernetes.io/control-plane
              operator: Exists
              effect: NoSchedule
        fileTailers:
          - name: kube-audit
            path: /var/lib/rancher/rke2/server/logs/*.log
      
      - name: workersnodesonly
        enabled: true
        workloadOverrides:
          nodeSelector:
            node-role.kubernetes.io/worker: "true"
        fileTailers:
          - name: kube-audit
            path: /var/lib/rancher/rke2/agent/logs/*.log
```

### DIFF
--- a/charts/logging-operator/values.yaml
+++ b/charts/logging-operator/values.yaml
@@ -284,6 +284,47 @@ logging:
     # -- containerOverrides
     containerOverrides:
 
+  # DEPRECATED: HostTailer config will be removed, use hostTailers instead
+  hostTailer:
+    # -- HostTailer
+    enabled: false
+    # -- name of HostTailer
+    name: hosttailer
+    image:
+      # -- repository of eventTailer image
+      repository:
+      # -- tag of eventTailer image
+      tag:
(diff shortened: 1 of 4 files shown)


charts/logging-operator/values.yaml
```diff
@@ -284,45 +284,46 @@ logging:
     # -- containerOverrides
     containerOverrides:
 
-  hostTailer:
```
csatib02, line 287: This should also be kept, with a deprecation notice

pepov (approved):

Issue #1140, opened 2022-11-28
Secret not found error not reported in Output status


aslafy-z:
**Describe the bug**:
If an Output references a wrong or inexistent secret, the error is not reported into the Output's status.

**Expected behaviour**:
`Output.status.active` is set to false, `Flow.status.active` is set to false. Output should reference the error message in its status, and Flow should get, if there was a single output configured in it, `status.active: false` too.

**Steps to reproduce the bug**:
- Create an Output that references a non existing secret.
- Create a Flow that references the previous Output.
- Check Output/Flow status

**Environment details**:
- logging-operator version (e.g. 2.1.1): 3.17.10

/kind bug

pepov:
Idea: the validation reconciler runs after the configuration is generated so in theory we can already depend on the loggingResources variable which contains all the outputs to also register and associate the error with the given outputs and let the validation reconciler report it in the status.

csatib02:
Revisited this, and i do see the output but there were a couple of errors regarding status reporting, opening a PR with a fix.

```yaml
apiVersion: logging.banzaicloud.io/v1beta1
kind: Output
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"logging.banzaicloud.io/v1beta1","kind":"Output","metadata":{"annotations":{},"name":"file","namespace":"logging"},"spec":{"datadog":{"api_key":{"valueFrom":{"secretKeyRef":{"key":"password","name":"secret"}}}}}}
  creationTimestamp: "2025-01-21T10:27:21Z"
  generation: 1
  name: file
  namespace: logging
  resourceVersion: "1280"
  uid: d2c5d5c0-30e2-4dcb-a962-aba45189199c
spec:
  datadog:
    api_key:
      valueFrom:
        secretKeyRef:
          key: password
          name: secret
status:
  active: true
  problems:
  - 'failed to get kubernetes secret logging:secret: Secret "secret" not found'
  problemsCount: 1
```

Issue #700, opened 2021-03-09
Orphan ClusterFlow and ClusterOutput resources are not reflected in their status


pepov:
The status of the ClusterFlow and ClusterOutput resources should highlight when they are not in the configured controlNamespace.

Solution idea:
collect all the ClusterFlows and ClusterOutputs with other logging resources that are not matching with the control namespace as well and then go through them in the validationreconiler to update their status.

pepov:
All flow and output resources should have their managing logging resource's name in their status, so that we can easily identify those resources that are orphan.

csatib02:
This issue was created a long time ago. Ever since then, the `validation reconciler` was introduced, so this issue was resolved.

Currently, if a `flow`/`clusterFlow` or `output`/`clusterOutput` has a completely empty `Status` field, it is considered orphaned, since if it would have been selected by a logging resource, the status fields would be filled out promptly.

PR #1920, opened 2025-01-21
fix(validation-reconciler): correct output and flow status report

csatib02:
Fixes: https://github.com/kube-logging/logging-operator/issues/1140
Fixes: https://github.com/kube-logging/logging-operator/issues/700

### DIFF
--- a/docs/configuration/crds/v1beta1/logging_types.md
+++ b/docs/configuration/crds/v1beta1/logging_types.md
@@ -10,7 +10,7 @@ LoggingSpec defines the desired state of Logging
 
 ### allowClusterResourcesFromAllNamespaces (bool, optional) {#loggingspec-allowclusterresourcesfromallnamespaces}
 
-Allow configuration of cluster resources from any namespace. Mutually exclusive with ControlNamespace restriction of Cluster resources 
+Allow configuration of cluster resources from any namespace. Mutually exclusive with ControlNamespace restriction of Cluster resources WARNING: Becareful when turning this on and off as it can result in some resources being orphaned. 
 
 
 ### clusterDomain (*string, optional) {#loggingspec-clusterdomain}
--- a/pkg/resources/model/reconciler.go
+++ b/pkg/resources/model/reconciler.go
@@ -120,13 +120,19 @@ func NewValidationReconciler(
 			}
 
 			for _, ref := range flow.Spec.GlobalOutputRefs {
-				if output := resources.Fluentd.ClusterOutputs.FindByName(ref); output != nil {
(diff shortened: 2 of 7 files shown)


pkg/resources/model/reconciler.go
```diff
@@ -183,15 +209,22 @@ func NewValidationReconciler(
...
+			flow.Status.ManagedBy = markManagedLoggingResources(&resources, flow, flow.Namespace, haveLoggingWithClusterWideAccess)
 
 			for _, ref := range flow.Spec.GlobalOutputRefs {
-				if output := resources.SyslogNG.ClusterOutputs.FindByName(ref); output != nil {
+				switch output := resources.SyslogNG.ClusterOutputs.FindByName(ref); {
+				case output == nil:
+					flow.Status.Problems = append(flow.Status.Problems, fmt.Sprintf("dangling global output reference: %s", ref))
+
+				case output.Status.ProblemsCount > 0:
+					flow.Status.Problems = append(flow.Status.Problems, fmt.Sprintf("Flow's: (%s) global output reference: %s has problems", flow.Name, output.Name))
```
pepov, line 220: not sure if we need the flow name, since we are writing the flow's status

docs/configuration/crds/v1beta1/logging_types.md
```diff
@@ -10,7 +10,7 @@ LoggingSpec defines the desired state of Logging
 
 ### allowClusterResourcesFromAllNamespaces (bool, optional) {#loggingspec-allowclusterresourcesfromallnamespaces}
 
-Allow configuration of cluster resources from any namespace. Mutually exclusive with ControlNamespace restriction of Cluster resources 
+Allow configuration of cluster resources from any namespace. Mutually exclusive with ControlNamespace restriction of Cluster resources WARNING: Becareful when turning this on and off as it can result in some resources being orphaned.
```
pepov, line 13: Is this still valid with the immutable constraint being present?

pepov (approved):

docs/configuration/crds/v1beta1/logging_types.md
csatib02: If a cluster output gets selected by a single logging, because of this flag being turned on, then its status will be filled out. If we turn this off the cluster output will not be selected, but its status will indicate that it is still active. So yeah users can make resource stuck if they turn this on and off.

pepov: the immutable check doesn't protect against removing an optional field, so the comment still apply

PR #1926, opened 2025-01-23
fix: add 'ruby-bigdecimal' to solve 'Oj is not installed, and falling…

sebastiangaiser:
… back to Yajl for json parser'

FluentD docs: https://docs.fluentd.org/quickstart/faq#fluentd-warns-oj-is-not-installed-and-falling-back-to-yajl-for-json-parser

Related to: https://github.com/kube-logging/fluentd-images/pull/184


pepov (approved):
Lets try this out here and we can merge for the legacy images as well!

PR #1929, opened 2025-01-27
Fine tune watch selectors for less memory usage

pepov:
The point is to introduce a backwards compatible option to reduce memory usage in large environments.

The PR introduces two new operator flags:
- `watch-labeled-children`: only watch child resources (pods, daemonsets, configmaps, etc) created by the operator with specific labels to avoid watching resources the operator is not concerned with. This option will be enabled by default in a future new **minor** version.
- `watch-labeled-secrets`: only watch secrets with `logging.banzaicloud.io/watch: enabled` label. This has the risk of breaking the operator if certain secrets are not labeled correctly. This option will be enabled by default in a future new **major** version.

Fixes: #1930 in https://github.com/kube-logging/logging-operator/pull/1929/commits/5f5aee023fc1890b5a6b782c061cff39c31b5736

### DIFF
--- a/main.go
+++ b/main.go
@@ -337,89 +337,92 @@ func setupCustomCache(mgrOptions *ctrl.Options, syncPeriod string, namespace str
 		mgrOptions.Cache.SyncPeriod = &duration
 	}
 
-	if namespace == "" && loggingRef == "" && !watchLabeledChildren {
-		return mgrOptions, nil
-	}
-
-	var namespaceSelector fields.Selector
-	var labelSelector labels.Selector
-	if namespace != "" {
-		namespaceSelector = fields.Set{"metadata.namespace": namespace}.AsSelector()
-	}
-	if loggingRef != "" {
-		labelSelector = labels.Set{"app.kubernetes.io/managed-by": loggingRef}.AsSelector()
-	}
-	if watchLabeledChildren {
-		if labelSelector == nil {
-			labelSelector = labels.NewSelector()
+	if namespace != "" || loggingRef != "" || watchLabeledChildren {
+		var namespaceSelector fields.Selector
+		var labelSelector labels.Selector
+		if namespace != "" {
+			namespaceSelector = fields.Set{"metadata.namespace": namespace}.AsSelector()
+			mgrOptions.Cache.DefaultNamespaces = map[string]cache.Config{
+				namespace: {
+					FieldSelector: namespaceSelector,
+				},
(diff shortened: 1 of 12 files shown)


main.go
```diff
@@ -31,11 +31,14 @@ import (
 	prometheusOperator "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/spf13/cast"
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/batch/v1"
```
csatib02, line 34: nit: could we use `batchv1` for this import alias?

csatib02 (approved):
LGTM!

OverOrion (approved):

main.go
```diff
@@ -98,6 +104,8 @@ func main() {
 	flag.BoolVar(&enableprofile, "pprof", false, "Enable pprof")
 	flag.StringVar(&namespace, "watch-namespace", "", "Namespace to filter the list of watched objects")
 	flag.StringVar(&loggingRef, "watch-logging-name", "", "Logging resource name to optionally filter the list of watched objects based on which logging they belong to by checking the app.kubernetes.io/managed-by label")
+	flag.BoolVar(&watchLabeledChildren, "watch-labeled-children", false, "Only watch child resources with logging operator's name label selector: app.kubernetes.io/name: fluentd|fluentbit|syslog-ng")
+	flag.BoolVar(&watchLabeledSecrets, "watch-labeled-secrets", false, "Only watch secrets with the following label selector: logging.banzaicloud.io/watch: enabled")
```
csatib02, line 108: We should write about somewhere, whether these two can be used at the same time, what's their effect (ORed or ANDed together) in that scenario etc.
Maybe it's time to open up a docs page about LO deployment arguments.

pkg/resources/fluentd/configsecret.go
```diff
@@ -21,9 +21,11 @@ import (
 
 	"emperror.dev/errors"
 	"github.com/cisco-open/operator-tools/pkg/reconciler"
-	"github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
+	util "github.com/cisco-open/operator-tools/pkg/utils"
```
csatib02, line 24: nit: no import alias needed here

csatib02 (approved):
LGTM!
Thanks for this feature!

Issue #1955, opened 2025-02-25
Make Fluent Bit’s default log volume /var/lib/containers optional


DinaGamalMahmoud:
- We are trying to deploy logging-operator on GKE Autopilot and we face an issue running Fluentbit because GKE autopilot doesn't allow access to `/var/lib/docker/containers `and we need to use `/var/log/containers` instead

- After checking logging operator configuration to override it , we couldn't because it is configured on Daemonset go script https://github.com/kube-logging/logging-operator/blob/master/pkg/resources/fluentbit/daemonset.go#L177 

 
- When I tried to change it and rebuild the Docker file it works as expected

- Can we have this change on the upcoming release

PR #1959, opened 2025-02-27
feat: add varlibcontainers volume as optional

DinaGamalMahmoud:
Fixes https://github.com/kube-logging/logging-operator/issues/1955

### DIFF
--- a/pkg/resources/fluentbit/daemonset.go
+++ b/pkg/resources/fluentbit/daemonset.go
@@ -177,11 +177,6 @@ func newConfigMapReloader(spec *v1beta1.FluentbitSpec) corev1.Container {
 func (r *Reconciler) generateVolumeMounts() (v []corev1.VolumeMount) {
 	v = []corev1.VolumeMount{
 		{
-			Name:      "varlibcontainers",
-			ReadOnly:  true,
-			MountPath: "/var/lib/docker/containers",
-		},
-		{
 			Name:      "varlogs",
 			ReadOnly:  true,
 			MountPath: "/var/log/",
@@ -190,6 +185,14 @@ func (r *Reconciler) generateVolumeMounts() (v []corev1.VolumeMount) {
 			Name:      "config",
 			MountPath: OperatorConfigPath,
 		},
+	}
+
+	if !*r.fluentbitSpec.DisableVarLibDockerContainers {
+		v = append(v, corev1.VolumeMount{
+			Name:      "varlibcontainers",
+			ReadOnly:  true,
+			MountPath: "/var/lib/docker/containers",
+		})
 	}
 
 	for vCount, vMnt := range r.fluentbitSpec.ExtraVolumeMounts {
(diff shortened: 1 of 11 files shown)


pkg/resources/fluentbit/daemonset.go
```diff
@@ -175,22 +175,27 @@ func newConfigMapReloader(spec *v1beta1.FluentbitSpec) corev1.Container {
 }
 
 func (r *Reconciler) generateVolumeMounts() (v []corev1.VolumeMount) {
-	v = []corev1.VolumeMount{
-		{
+	v = []corev1.VolumeMount{}
+
+	if r.fluentbitSpec.EnableVarLibContainers == nil || *r.fluentbitSpec.EnableVarLibContainers {
```
csatib02, line 180: So I would suggest using `r.fluentbitSpec.DisableVarLibContainers` the conditional could be `*r.fluentbitSpec.DisableVarLibContainers` and in [here](https://github.com/kube-logging/logging-operator/blob/e0670708ff2008bd0f8957f134ba08e456f1bf16/pkg/sdk/logging/api/v1beta1/logging_types.go#L268) you can set a default to avoid a nil pointer.

pkg/resources/fluentbit/daemonset.go
```diff
@@ -175,22 +175,27 @@ func newConfigMapReloader(spec *v1beta1.FluentbitSpec) corev1.Container {
...
 			ReadOnly:  true,
 			MountPath: "/var/log/",
 		},
-		{
+		corev1.VolumeMount{
 			Name:      "config",
 			MountPath: OperatorConfigPath,
 		},
-	}
+	)
```
csatib02, line 198: Since these are added regardless, we should keep it as it was, meaning that the slice is initialized with this.

pkg/resources/fluentbit/daemonset.go
```diff
@@ -201,36 +206,39 @@ func (r *Reconciler) generateVolumeMounts() (v []corev1.VolumeMount) {
...
+		corev1.Volume{
 			Name: "varlogs",
 			VolumeSource: corev1.VolumeSource{
 				HostPath: &corev1.HostPathVolumeSource{
 					Path: "/var/log",
 				},
 			},
 		},
-	}
+	)
```
csatib02, line 241: Same as before, initialize the slice with this.

csatib02 (changes requested):

pkg/sdk/logging/api/v1beta1/fluentbit_types.go
```diff
@@ -108,6 +108,7 @@ type FluentbitSpec struct {
 	FilterKubernetes FilterKubernetes `json:"filterKubernetes,omitempty"`
 	// Disable Kubernetes metadata filter
 	DisableKubernetesFilter *bool         `json:"disableKubernetesFilter,omitempty"`
+	EnableVarLibContainers  *bool         `json:"enableVarLibContainers,omitempty"`
```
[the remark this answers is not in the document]
aslafy-z: Shouldn't this be `DisableVarLibDockerContainers`?

csatib02 (approved):
LGTM!

Issue #1987, opened 2025-03-19
Run fluentd as user by default


pepov:
**Is your feature request related to a problem? Please describe.**
Fluentd currently runs with root which is unreasonable from a security perspective.

**Describe the solution you'd like**
Fluentd should run as the fluentd user which is already configured in the image.

**Describe alternatives you've considered**
Make this a documented option, but leave it as is.

**Additional context**
We need to consider if this might break existing deployments, but hopefully it shouldn't.

PR #2002, opened 2025-04-02
feat: set default user and group

csatib02:
Fixes: #1987


pepov:
Relabeled it as a breaking change, so that users who depended on the previous behaviour somehow will take attention

pepov (approved):

csatib02:
@sebastiangaiser 
I believe yes, since these are just the defaults. If user input was given for a defaulted value it should be overwritten, if a value was not defaulted then it should be added. Have you checked this?

sebastiangaiser:
It gets overwritten (LO - 5.3.0):
```
  securityContext:
    fsGroup: 101
    seccompProfile:
      type: RuntimeDefault
```

csatib02:
Yeah this check currently works only if the SecurityContext or in your case PodSecurityContext is nil.
We can add some logic to merge this, but as for this PR it works as intended.

PR #2032, opened 2025-05-02
fix(helm): priorityClassName must be of type string

### LINKED ISSUE #2031: Helm: Default value for priorityClassName is a map not string
**Describe the bug**:
Setting the `priorityClassName` in the helm values, results in the following error if the Chart is used as dependency.

```
coalesce.go:298: warning: cannot overwrite table with non table for priorityClassName (map[])
```

**Expected behaviour**:
Setting the `priorityClassName` to a string should work even with more recent Helm versions.

Use the following value.yaml

```
priorityClassName: "existing-priority-class"
```

**Additional context**:

This seems to be a new behavior introduced with helm/helm#12469 and present since Helm v3.17 release.

**Environment details**:
- logging-operator version (e.g. 2.1.1): 5.3.0
- Helm version: v3.17+

/kind bug

peschmae:
The default value of `priorityClassName` is currently a map/object, but is used as a string when setting it on the deployment.

This change sets the default value to null (using `~`), and is now considered a string value.

Fixes #2031


csatib02 (approved):
Thanks!

PR #2042, opened 2025-05-18
feat: add compress field to forward output

cyril-corbon:
add compress field to forward output
https://docs.fluentd.org/output/forward#how-to-enable-gzip-compression


OverOrion (approved):
Thank you for your contriubution @cyril-corbon! :rocket:

PR #2043, opened 2025-05-20
feat: add parameter to disable mount of /var/log directory

volleynator:
I want to make the mount of /var/log optional as there might be users of k8s clusters which do not allow mounting host volumes but provide different ways to collect host logs.

### DIFF
--- a/pkg/resources/fluentbit/daemonset.go
+++ b/pkg/resources/fluentbit/daemonset.go
@@ -176,14 +176,17 @@ func newConfigMapReloader(spec *v1beta1.FluentbitSpec) corev1.Container {
 func (r *Reconciler) generateVolumeMounts() (v []corev1.VolumeMount) {
 	v = []corev1.VolumeMount{
 		{
-			Name:      "varlogs",
-			ReadOnly:  true,
-			MountPath: "/var/log/",
-		},
-		{
 			Name:      "config",
 			MountPath: OperatorConfigPath,
 		},
+	}
+
+	if !*r.fluentbitSpec.DisableVarLog {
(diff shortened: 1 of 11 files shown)


csatib02 (commented):
LGTM!

Also please sign your commits, there are instructions if you click on the failing check.

pkg/resources/fluentbit/daemonset.go
```diff
@@ -174,16 +174,20 @@ func newConfigMapReloader(spec *v1beta1.FluentbitSpec) corev1.Container {
...
+		v = []corev1.VolumeMount{
+			{
+				Name:      "varlogs",
+				ReadOnly:  true,
+				MountPath: "/var/log/",
+			},
+			{
+				Name:      "config",
+				MountPath: OperatorConfigPath,
+			},
```
aslafy-z, line 189: I think this should move back to the default volume mount list at line 177

csatib02: Yes it should, nice catch!

pkg/resources/fluentbit/daemonset.go
```diff
@@ -174,16 +174,20 @@ func newConfigMapReloader(spec *v1beta1.FluentbitSpec) corev1.Container {
...
+		}
```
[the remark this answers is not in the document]
volleynator: fixed that

volleynator:
@csatib02 thx for the feedback. I signed the commit, DCO check succeeds now.

csatib02 (approved):

aslafy-z (approved):
LGTM!

PR #2044, opened 2025-05-21
fix version 5.3.0 instead of 0.0.0

jimmytuc:
The version `0.0.0` is a bit weird and somehow it stopped the fluentd to use the right ES fluentd plugin https://github.com/uken/fluent-plugin-elasticsearch version `6.0.0` https://github.com/uken/fluent-plugin-elasticsearch/pull/1067

Specifying the fixed version seems solved the issue in our end

### DIFF
--- a/charts/logging-operator/Chart.yaml
+++ b/charts/logging-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 type: application
 name: logging-operator
-version: 0.0.0
-appVersion: latest
+version: 5.3.0
+appVersion: 5.3.0
 kubeVersion: ">=1.22.0-0"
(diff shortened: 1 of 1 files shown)


sagikazarmark:
The chart version is updated during the release process. There is no need to update this version manually.

pepov:
If you are using the chart source you have the ability to change the version of the chart right? Please try to solve this in your custom process or controbute your recommended chart changes upstream but we definitely don't want to have any hardcoded version of the logging operator or the chart itself in the code as it would complicate our release process. I hope this sounds reasonable!

csatib02:
Closing this, based on the above responses.

If you happened to have anymore questions, feel free open a ticket!

PR #2046, opened 2025-05-29
chore: unify metrics probe keys name in axosyslog

OverOrion:


pkg/sdk/logging/model/syslogng/config/log.go
```diff
@@ -74,7 +74,7 @@ func destinationLogPath(dest destination) render.Renderer {
 		}
 		if v, ok := m.Labels["destination"]; !ok || v != "" {
```
furiel, line 75: You need to update here too.

pepov (approved):

PR #2059, opened 2025-06-05
Add support for Logicmonitor

MaxAnderson95:
This PR adds support for the [`lm-logs-fluentd`](https://github.com/logicmonitor/lm-logs-fluentd?tab=readme-ov-file) plugin for use with Logicmonitor.

I tried to match how the other plugins were implemented, but Go is not a language I work with a lot so forgive me if there are some things that need to be adjusted.

I also updated the Docker images to include the `lm-logs-fluentd` plugin.

Please let me know if any adjustments are needed!


csatib02 (approved):
LGTM!

Great work!

OverOrion (approved):
Thank you for your contribution!

PR #2247, opened 2026-06-04
fix: add Fluentd configuration template

KateFiroozi:
`logging-operator` Helm charts is missing template file for `FluentdConfig` resource. `Values.yaml` includes configuration for the resource, but since template is missing, it will be ignored and no resource created.


promptless-for-oss:
Promptless prepared a documentation update related to this change.

Triggered by [PR #2247](https://github.com/kube-logging/logging-operator/pull/2247)

This fix adds the missing FluentdConfig Helm template, enabling users to configure Fluentd through Helm values. The documentation update describes how to use `logging.fluentd` values and the conditions for FluentdConfig resource creation.

Review: [Document FluentdConfig Helm template configuration](https://github.com/kube-logging/kube-logging.github.io/pull/334)

charts/logging-operator/templates/logging/fluentd.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.logging.enabled (not .Values.logging.fluentdDisabled) -}}
+{{- $fluentdSpec := .Values.logging.fluentd }}
+{{- if .Values.logging.loggingRef }}
+  {{- $fluentdSpec := set .Values.logging.fluentd "loggingRef" (default .Values.logging.loggingRef .Values.logging.fluentd.loggingRef) -}}
+{{- end }}
```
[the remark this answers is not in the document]
KateFiroozi: template for FluentdConfig is based (cpoied from) template for FluentbitAgent

charts/logging-operator/templates/logging/fluentd.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.logging.enabled (not .Values.logging.fluentdDisabled) -}}
```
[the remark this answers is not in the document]
KateFiroozi: fixed in https://github.com/kube-logging/logging-operator/pull/2247/commits/62daa5d48f0f9b403b67f6c72cbf9f5306cb9919

csatib02:
Thanks for this @KateFiroozi!

One blocker before merge: this duplicates the fluentd config path.

`templates/logging/logging.yaml` already renders `.Values.logging.fluentd` **inline** as `spec.fluentd` on the `Logging` resource. This new template reads the **same value** under the **same conditions** and emits it again as a detached `FluentdConfig`. So when a user sets `logging.fluentd`, the chart now produces the spec twice.
The operator treats this as a conflict (`pkg/resources/model/reconciler.go:284-290`) and raises a persistent problem.
Note `fluentbit.yaml` doesn't hit this because `logging.yaml` has no inline `fluentbit:` block — fluentbit was already fully migrated to detached-only. Fluentd was only half-migrated here.

Please remove the inline fluentd block in `logging.yaml` so the detached `FluentdConfig` is the single source of truth (matches the fluentbit pattern). This is a behavior change for existing users, so worth a note in the chart changelog. Alternatively, render from a dedicated key like `.Values.logging.fluentdConfig` to keep the two paths separate.

charts/logging-operator/templates/logging/fluentd.yaml
```diff
@@ -0,0 +1,17 @@
...
+{{- $fluentdSpec := .Values.logging.fluentd }}
+{{- if .Values.logging.loggingRef }}
+  {{- $fluentdSpec := set .Values.logging.fluentd "loggingRef" (default .Values.logging.loggingRef .Values.logging.fluentd.loggingRef) -}}
```
csatib02: This inner `$fluentdSpec` is block-scoped and discarded at `end`. The outer var used below is never reassigned. `loggingRef` only lands via the `set` side-effect mutating the map in place, which is fragile. Also `.Values.logging.fluentd.loggingRef` can nil-deref if `fluentd` is `null`.

```suggestion
{{- $fluentdSpec := .Values.logging.fluentd | default dict -}}
{{- if .Values.logging.loggingRef }}
  {{- $fluentdSpec = set $fluentdSpec "loggingRef" (default .Values.logging.loggingRef $fluentdSpec.loggingRef) -}}
{{- end }}
```

KateFiroozi:
Thank for the feedback, @csatib02 ! I will take this into account and will update this PR.

KateFiroozi:
@csatib02 , I have made it optional to create a separate resource or to follow established way, so no breaking changes for the users.

charts/logging-operator/templates/logging/fluentd.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if and .Values.logging.enabled (not .Values.logging.fluentdDisabled) .Values.logging.fluentdConfig.create -}}
+{{- $fluentdSpec := .Values.logging.fluentd }}
+{{- if .Values.logging.loggingRef }}
+{{- $fluentdSpec := .Values.logging.fluentd | default dict -}}
+{{- if .Values.logging.loggingRef }}
+  {{- $fluentdSpec = set $fluentdSpec "loggingRef" (default .Values.logging.loggingRef $fluentdSpec.loggingRef) -}}
+{{- end }}
+{{- end }}
```
csatib02: This block declares a **new** `$fluentdSpec` scoped to the outer `if`, the line assignment after that mutates *that* one, which is discarded at `end`.

With `loggingRef` set and `fluentd` empty (`{}`, the default), the rendered `FluentdConfig` is `spec: {}` — **no `loggingRef`**.

checked with:

```bash
helm template . --set logging.enabled=true \
    --set logging.fluentdConfig.create=true --set logging.loggingRef=myref
# FluentdConfig renders: spec: {}   <-- loggingRef missing
```


```suggestion
{{- $fluentdSpec := .Values.logging.fluentd | default dict -}}
{{- if .Values.logging.loggingRef }}
  {{- $fluentdSpec = set $fluentdSpec "loggingRef" (default .Values.logging.loggingRef $fluentdSpec.loggingRef) -}}
{{- end }}
```

charts/logging-operator/values.yaml
```diff
@@ -166,6 +166,11 @@ logging:
 
   # -- Flag to disable fluentd completely
   fluentdDisabled: false
+  # -- Create a FluentdConfig as a separate custom resource.
+  # Default is false, which means that the fluentd configuration will be generated as part of the Logging custom resource if fluentd is enabled.
+  # Setting this to true will create a separate FluentdConfig custom resource with the same name as the Logging resource and reference it in the Logging.fluentd section.
```
csatib02: nit:

Setting this to true will create a separate FluentdConfig custom resource with the same name as the Logging resource and **reference it in the Logging.fluentd section.**

The template doesn't add any reference in `Logging.fluentd` — it just omits the inline block, and association happens by matching name/`loggingRef`.

```suggestion
  # Setting this to true will create a separate FluentdConfig custom resource with the same name as the Logging resource.
```

csatib02 (commented):
One minor blocker and we are good to go! :)

csatib02 (approved):
LGTM!

csatib02:
@KateFiroozi, please run `make generate` or `make check-diff` and commit the helm-docs changes please.

KateFiroozi:
@csatib02 , done!

promptless-for-oss:
Promptless updated the documentation for this change.

Triggered by [PR #2247](https://github.com/kube-logging/logging-operator/pull/2247) merge

The documentation has been revised to reflect the final implementation: the `logging.fluentdConfig.create` flag is opt-in (default `false`), and when enabled, creates a separate FluentdConfig resource while omitting the inline fluentd configuration from the Logging resource.

Review: [Document FluentdConfig Helm chart template](https://github.com/kube-logging/kube-logging.github.io/pull/334)

katerynafiroozi:
@csatib02 , a quick question. I see changes has been merged into master, but will new version of chart released later? By pulling 6.5.2 version from the registry, it does not include changes present in a master branch

csatib02:
@katerynafiroozi, The chart only gets published on releases, which we will do in 2 weeks approximately.
For the time being you may use the latest version from the CI artifacts: https://github.com/kube-logging/logging-operator/actions/runs/27102256204

PR #2267, opened 2026-07-09
feat(syslog-ng): expose log_msg_size global option

fdolsky321:
## Summary

Exposes syslog-ng's `log-msg-size()` as a configurable global option so users can override the built-in default of **8192** bytes for larger log records.

- Adds `log_msg_size` (`*int`) to `SyslogNGSpec.GlobalOptions` in `pkg/sdk/logging/api/v1beta1/syslogng_types.go`.
- The reflection-based config renderer emits `log_msg_size(<n>);` inside the top-level `options { }` block. When unset, nothing is rendered and syslog-ng keeps its native 8192 default (no behavior change for existing users).
- Regenerated deepcopy, CRD manifests (Logging + SyslogNGConfig), Helm chart CRDs, and docs via `make generate`.

## Usage

```yaml
spec:
  syslogNG:
    globalOptions:
      log_msg_size: 32768
```

## Testing

- Added a golden-config test case (`global options log_msg_size`) in `pkg/sdk/logging/model/syslogng/config/config_test.go` asserting `log_msg_size(32768);` renders in the `options { }` block.
- `go build ./...` and `go test ./logging/model/syslogng/...` pass in `pkg/sdk`.


OverOrion (approved):

OverOrion:
Thank you for your contribution @fdolsky321!

PR #2273, opened 2026-07-20
fix: enable dual-stack for Fluent Bit metrics services

### LINKED ISSUE #2238: FluentBit metrics service scraping over IPv6
We want to use IPv6 for as many connections as possible in our cluster. We have set `enabledIPv6: true` in our `FluentbitAgent` resource, however the operator-managed `fluentbit-metrics` service is still being rendered as `SingleStack`.

It would be awesome if it would be `PreferDualStack` instead of `SingleStack` when `enabledIPv6: true` is set.

A mutation admission policy that modifies the service

**Additional context**
Similar request was in https://github.com/kube-logging/logging-operator/issues/2100 previously and I think the PR https://github.com/kube-logging/logging-operator/pull/2117 just missed this?

vyncint:
## Summary

- configure Fluent Bit metrics and buffer metrics Services for dual-stack networking when `enabledIPv6` is enabled
- preserve the existing IP family behavior when `enabledIPv6` is disabled
- add focused unit coverage for both modes

## Testing

- `make check`
- `make check-diff`

Fixes #2238

### DIFF
--- a/pkg/resources/fluentbit/service.go
+++ b/pkg/resources/fluentbit/service.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/kube-logging/logging-operator/pkg/resources/model"
+	"github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
 )
 
 func (r *Reconciler) serviceMetrics() (runtime.Object, reconciler.DesiredState, error) {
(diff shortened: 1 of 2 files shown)


promptless-for-oss:
📝 **Promptless** drafted a documentation update for this change.

We've prepared a "What's new" entry documenting that setting `enabledIPv6: true` now also enables dual-stack networking for the Fluent Bit metrics and buffer-volume metrics Services, so those metrics endpoints are reachable over IPv6 on dual-stack clusters — matching the behavior already documented for Fluentd and syslog-ng.

Because this change hasn't shipped in a release yet, the note is currently filed under an "Unreleased" heading; the version heading should be updated to the actual release once this PR merges and is cut into a release.

Review the suggestion here: https://app.gopromptless.ai/suggestions/fac15125-2213-4ded-87c3-444ad974009b

csatib02 (approved):

csatib02:
Thanks @vyncint!

PR #2282, opened 2026-07-26
feat(fluent-bit): expose storage.backlog.flush_on_shutdown option

### LINKED ISSUE #2176: Support the flush_on_shutdown option
When running FluentBit on transient infrastructure like EC2 spot instances, logs can be permanently lost when instances are terminated. FluentBit buffers logs to disk, but without the `storage.backlog.flush_on_shutdown` configuration enabled, the buffer is not flushed during graceful shutdown. This results in data loss when spot instances are reclaimed or nodes are drained, as the buffered logs are deleted along with the instance.

Add support for the `storage.backlog.flush_on_shutdown` configuration parameter in the `BufferStorage` CRD spec. This would allow users to configure FluentBit to flush buffered logs during graceful shutdown, preventing data loss on transient infrastructure.

The configuration is already supported in FluentBit's service section ([documentation](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/yaml/service-section#storage-configuration)) but is not exposed in the Logging Operator's FluentBit CRD ([current spec](https://kube-logging.dev/docs/configuration/crds/v1beta1/fluentbit_types/#bufferstorage)).

Proposed addition to `BufferStorage`:
```yaml
kind: FluentbitAgent
spec:
  bufferStorage:
    storage.backlog.flush_on_shutdown: true  # new field
```

- **Increasing termination grace periods**: This only delays the problem and doesn't guarantee log delivery
- **External FluentBit configuration**: Managing FluentBit outside the operator defeats the purpose of using the Logging Operator

**Additional context**
This is particularly critical for:
- Spot instance/preemptible VM workloads
- Kubernetes clusters with frequent node rotations
- Cost-optimized infrastructure with transient compute resources

The FluentBit project already provides this safety mechanism; the Logging Operator just needs to expose it through the CRD.

vyncint:
## Summary

Fluent Bit can flush filesystem backlog chunks to their destination during graceful shutdown via `storage.backlog.flush_on_shutdown`, but the option was not reachable through the `BufferStorage` spec. Without it, buffered records are lost when a node is drained or a spot instance is reclaimed — the case described in #2176.

Closes #2176

## Changes

- `pkg/sdk/logging/api/v1beta1/fluentbit_types.go`: add `StorageBacklogFlushOnShutdown` to `BufferStorage`, mapped to `storage.backlog.flush_on_shutdown`.
- Regenerated CRDs, chart CRDs and `docs/configuration/crds/v1beta1/fluentbit_types.md` with `make manifests docs`.
- `pkg/resources/fluentbit/configsecret_test.go`: render test covering enabled, explicitly disabled, and unset.

```yaml
kind: FluentbitAgent
spec:
  bufferStorage:
    storage.path: /buffers
    storage.backlog.flush_on_shutdown: "On"
```

renders into the `[SERVICE]` section as:

```
    storage.path  /buffers
    storage.backlog.flush_on_shutdown  On
```

## Notes

The field is a `string` rather than a `*bool`, matching the neighbouring on/off toggles in the same struct (`storage.checksum`, `storage.delete_irrecoverable_chunks`, `storage.metrics`) and Fluent Bit's own `off`/`on` values. The issue sketched `true`; happy to switch to `*bool` if you would rather the API read that way, though it would then be the only boolean in `BufferStorage`.

It needs no template change: `BufferStorage` already reaches the `[SERVICE]` section through `StructToStringMapper`, and `omitempty` plus the template's `{{- if $value }}` guard keep it out of the rendered config when unset.

## Verification

- `go test ./pkg/...` — all packages pass, including the new `TestBufferStorageServiceSection` (3 subtests: enabled, explicitly `Off`, and unset staying absent from the config).
- `cd pkg/sdk/logging && go test ./...` with envtest assets — passes.
- `make lint` — 0 issues in both modules.
- `make generate` run a second time produces no further diff, so `check-diff` stays green.
- Option name, accepted values (`off`/`on`) and default (`off`) checked against the Fluent Bit service-section documentation.

### DIFF
--- a/charts/logging-operator/charts/logging-operator-crds/templates/logging.banzaicloud.io_fluentbitagents.yaml
+++ b/charts/logging-operator/charts/logging-operator-crds/templates/logging.banzaicloud.io_fluentbitagents.yaml
@@ -478,6 +478,8 @@ spec:
                 type: object
               bufferStorage:
                 properties:
+                  storage.backlog.flush_on_shutdown:
+                    type: string
                   storage.backlog.mem_limit:
                     type: string
                   storage.checksum:
--- a/charts/logging-operator/charts/logging-operator-crds/templates/logging.banzaicloud.io_loggings.yaml
+++ b/charts/logging-operator/charts/logging-operator-crds/templates/logging.banzaicloud.io_loggings.yaml
@@ -1339,6 +1339,8 @@ spec:
                     type: object
                   bufferStorage:
                     properties:
+                      storage.backlog.flush_on_shutdown:
+                        type: string
                       storage.backlog.mem_limit:
                         type: string
                       storage.checksum:
--- a/charts/logging-operator/crds/logging.banzaicloud.io_fluentbitagents.yaml
+++ b/charts/logging-operator/crds/logging.banzaicloud.io_fluentbitagents.yaml
@@ -475,6 +475,8 @@ spec:
                 type: object
               bufferStorage:
                 properties:
+                  storage.backlog.flush_on_shutdown:
+                    type: string
                   storage.backlog.mem_limit:
                     type: string
                   storage.checksum:
--- a/charts/logging-operator/crds/logging.banzaicloud.io_loggings.yaml
+++ b/charts/logging-operator/crds/logging.banzaicloud.io_loggings.yaml
@@ -1336,6 +1336,8 @@ spec:
                     type: object
                   bufferStorage:
                     properties:
+                      storage.backlog.flush_on_shutdown:
+                        type: string
                       storage.backlog.mem_limit:
                         type: string
                       storage.checksum:
--- a/config/crd/bases/logging.banzaicloud.io_fluentbitagents.yaml
+++ b/config/crd/bases/logging.banzaicloud.io_fluentbitagents.yaml
@@ -475,6 +475,8 @@ spec:
                 type: object
               bufferStorage:
                 properties:
+                  storage.backlog.flush_on_shutdown:
+                    type: string
                   storage.backlog.mem_limit:
                     type: string
                   storage.checksum:
--- a/config/crd/bases/logging.banzaicloud.io_loggings.yaml
+++ b/config/crd/bases/logging.banzaicloud.io_loggings.yaml
@@ -1336,6 +1336,8 @@ spec:
                     type: object
                   bufferStorage:
                     properties:
+                      storage.backlog.flush_on_shutdown:
+                        type: string
                       storage.backlog.mem_limit:
                         type: string
                       storage.checksum:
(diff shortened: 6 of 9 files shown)


vyncint:
Verified end to end on a throwaway kind cluster, since unit tests alone cannot show that the API server accepts a dotted property name or that the pinned Fluent Bit build recognises the option.

1. The CRD admits the field and round-trips it. Applied a `FluentbitAgent` against the regenerated CRD and read it back:

```
$ kubectl get fluentbitagent flush-on-shutdown-check -o jsonpath='{.spec.bufferStorage}'
{"storage.backlog.flush_on_shutdown":"On","storage.backlog.mem_limit":"10M","storage.path":"/buffers"}
```

Negative control, same manifest with the key misspelled, is rejected — so the acceptance above is meaningful and not just a permissive schema:

```
Error from server (BadRequest): error when creating "agent-bogus.yaml": FluentbitAgent in version
"v1beta1" cannot be handled as a FluentbitAgent: strict decoding error: unknown field
"spec.bufferStorage.storage.backlog.flush_on_shutdown_typo"
```

2. The operator writes it into the generated config. Built the controller image from this branch, loaded it into kind, installed the chart from this branch, then read the generated secret:

```
$ kubectl -n logging get secret e2e-fluentbit -o jsonpath='{.data.fluent-bit\.conf}' | base64 -d

[SERVICE]
    Flush        1
    Grace        5
    Daemon       Off
    Log_Level    info
    Parsers_File /fluent-bit/etc/parsers.conf
    Coro_Stack_Size    24576
    storage.backlog.flush_on_shutdown  On
    storage.backlog.mem_limit  10M
    storage.path  /buffers
```

3. The pinned Fluent Bit accepts it at runtime. The DaemonSet pod runs the default `ghcr.io/fluent/fluent-bit:5.0.5` with that config:

```
$ kubectl -n logging get pods -l app.kubernetes.io/name=fluentbit
NAME                  READY   STATUS    RESTARTS   AGE
e2e-fluentbit-qs4fs   1/1     Running   0          36s
```

No `unknown`, `unrecognized` or `invalid configuration` lines in its log; the only `storage.backlog` entries are the normal informational ones:

```
[ info] [storage] backlog input plugin: storage_backlog.1
[ info] [input:storage_backlog:storage_backlog.1] initializing
[ info] [input:storage_backlog:storage_backlog.1] storage_strategy='memory' (memory only)
[ info] [input:storage_backlog:storage_backlog.1] queue memory limit: 9.5M
```

Two notes on method. `fluent-bit --dry-run` is not a useful check here: it reports `configuration test is successful` for a deliberately misspelled service key too, so it cannot tell a recognised option from an ignored one. What does discriminate is the binary itself — `storage.backlog.flush_on_shutdown` is present in the 5.0.5 build alongside `storage.backlog.mem_limit` and `storage.checksum`, while the misspelled variant is absent. Separately, the `no upstream connections available` errors in the pod log are just the forward output with no aggregator in this minimal setup, unrelated to this change.

csatib02 (approved):

PR #2284, opened 2026-07-27
docs(plugins): fix dedot description from being incorrectly described as concat

halkeye:


csatib02 (approved):

halkeye:
looks like it is generated and pulled from somewhere. Tomorrow when i'm at a computer i'll track down from where

csatib02:
Yeah, in the repo just run `make generate`

halkeye:
Sorry about that, I just tried to make a quick change via github ui last night

source code is updated, make generate ran.

should be good now

PR #2286, opened 2026-07-27
fix(configcheck): merge aggregator level configCheck field by field

### LINKED ISSUE #2228: Fluentd ConfigCheck pod misses PodSecurityContext if configCheck is set in Logging resource
**Describe the bug**:
When you add
```
spec:
  configCheck:
    timeoutSeconds: 10
```
in the Logging custom resource, then the code in [newCheckPod](https://github.com/kube-logging/logging-operator/blob/master/pkg/resources/fluentd/appconfigmap.go#L264) is somehow not executed which results in a fluentd-checkconfig pod that misses certain elements.
Many of them may be missing but one of them is the PodSecurityContext, which, in combination with the configuration workers>1 in the FluentdConfig causes the configcheck to crash with the error described in #1023 

**Expected behaviour**:
If the mentioned `configCheck` configuration in Logging is added, then the runtime pod configuration of the fluentd-configcheck pod should still contain the configured pod spec. For example PodSecurityContext.

Add the mentioned `configCheck` configuration in Logging resource; also set some pod spec in the FluentdConfig resource, e.g. .spec.security.podSecurityContext.runAsUser = 1001.

**Additional context**:
The bug was discovered after analysis regarding the problem described in #1023 

**Environment details**:
- Kubernetes version (e.g. v1.15.2): 1.33.9
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): on premise Openshift
- logging-operator version (e.g. 2.1.1): 6.2.2
- Install method (e.g. helm or static manifests): static manifests
- Logs from the misbehaving component (and any other relevant logs): not necessary, just run kubectl get pod ...-fluentd-configcheck-... -oyaml
```
# Logging:
spec:
  ...
  configCheck:
    timeoutSeconds: 10
  ...
# FluentdConfig:
spec:
  ...
  security:
    podSecurityContext:
      runAsUser: 1001
  workers: 8
```

/kind bug

vyncint:
### Description

`AggregatorLevelConfigCheck` replaces the whole `ConfigCheck` struct:

```go
func (l *Logging) AggregatorLevelConfigCheck(check *ConfigCheck) {
	if check != nil {
		l.Spec.ConfigCheck = *check
		l.configCheckDefaults()
	}
}
```

So if a `FluentdConfig` or `SyslogNGConfig` sets *any* `configCheck` field, every field set on the `Logging` resource is silently dropped. With this on the `Logging`:

```yaml
configCheck:
  timeoutSeconds: 42
  labels:
    from: logging-level
fluentd:
  configCheck:
    strategy: StartWithTimeout   # the aggregator only overrides the strategy
```

`labels` is discarded and `timeoutSeconds` falls back to the default of 10. Both are user visible: `ConfigCheck.Labels` feeds `configCheckPodObjectMeta`, so the configcheck pod loses labels that were explicitly asked for, and `StartWithTimeout` runs with a timeout the user did not configure.

This PR overrides each field on its own. That matches what the `LoggingSpec.ConfigCheck` doc comment already promises — *"Can be overridden on the fluentd / syslog-ng level"* — and it is the same shape as the image spec fix in #2283. An aggregator that sets nothing, or sets an empty struct, now leaves the logging level settings alone.

I also updated the doc comments on the two aggregator side `ConfigCheck` fields, which described the old whole-struct copy and would otherwise contradict the code. Those regenerate into `docs/`; the CRDs are unaffected because the project builds them with `crd:maxDescLen=0`.

### Link to the issue in case of a bug fix.

Refs #2228

That issue reports the configcheck pod missing configured settings. I found this while investigating it and [wrote it up there](https://github.com/kube-logging/logging-operator/issues/2228#issuecomment-5091371764), offering a patch — this is that patch. It is independent of #2285, which fixes the `dnsPolicy` gap in the same area, so the two can land in either order. Happy to close this if you would rather keep the current replace semantics.

### Testing details

1. Manual - verified on a kind cluster (v1.33.1), A/B against the released operator, using the `Logging` above:

   | | configcheck pod label `from` | configcheck container args |
   |---|---|---|
   | released 6.7.0 | *(absent)* | `["timeout","10","fluentd",...]` |
   | this PR | `logging-level` | `["timeout","42","fluentd",...]` |

   The `Logging` was deleted and recreated between runs so the result was not served from `status.configCheckResults`, and I confirmed the "after" pod had a new UID and a `creationTimestamp` later than the operator rollout. The check still passed (`configCheckResults: {"893225a7": true}`, no `status.problems`) and the aggregator StatefulSet came up.

2. Unit tests - added `TestAggregatorLevelConfigCheck`, a table over nil / empty / strategy-only / timeout-only / labels-only / all-fields, and `TestAggregatorLevelConfigCheckDefaultsTimeout` pinning the fallback to 10 when neither level sets one.

   As a negative control I restored the whole-struct assignment and confirmed 4 of the 6 cases fail. The nil and all-fields cases pass either way by design — they are the boundaries where replace and per-field merge agree.

3. `go test ./pkg/... ./controllers/...` passes with envtest assets, `make lint` reports 0 issues in both modules, and `make generate` produces no diff beyond the files in this PR.

### Any backward incompatible change? If so, please explain.

Yes in principle, and it is worth an explicit decision.

An aggregator level `configCheck` that previously blanked a logging level field now leaves it in place. Anything relying on that blanking sees a change — but the only way to depend on it was to set a field on the `Logging` resource and expect it to be ignored, which is the bug being reported. Installations that set `configCheck` in only one place, or set every field in both, are unaffected.

There is one judgement call I want to flag rather than bury: the aggregator side doc comment used to say the field is *"just copied over the field in the logging resource if defined"*, which does describe the old behaviour. My reading is that the `LoggingSpec` side comment is the intended contract and `Labels` vanishing was never deliberate, but if you disagree, the alternative is to keep replace semantics and document it clearly instead.

### DIFF
--- a/docs/configuration/crds/v1beta1/fluentd_types.md
+++ b/docs/configuration/crds/v1beta1/fluentd_types.md
@@ -39,7 +39,7 @@ BufferStorageVolume is by default configured as PVC using FluentdPvcSpec [volume
 
 ### configCheck (*ConfigCheck, optional) {#fluentdspec-configcheck}
 
-Overrides the default logging level configCheck setup This field is not used directly, its fields are copied over the ones in the logging resource, field by field, so setting one field here does not discard the others set on the logging resource 
+Overrides the default logging level configCheck setup. This field is not used directly, its fields are copied over the ones in the logging resource, field by field, so setting one field here does not discard the others set on the logging resource. 
 
 
 ### configCheckAnnotations (map[string]string, optional) {#fluentdspec-configcheckannotations}
--- a/docs/configuration/crds/v1beta1/fluentd_types.md
+++ b/docs/configuration/crds/v1beta1/fluentd_types.md
@@ -39,7 +39,7 @@ BufferStorageVolume is by default configured as PVC using FluentdPvcSpec [volume
 
 ### configCheck (*ConfigCheck, optional) {#fluentdspec-configcheck}
 
-Overrides the default logging level configCheck setup This field is not used directly, just copied over the field in the logging resource if defined 
+Overrides the default logging level configCheck setup This field is not used directly, its fields are copied over the ones in the logging resource, field by field, so setting one field here does not discard the others set on the logging resource 
 
 
 ### configCheckAnnotations (map[string]string, optional) {#fluentdspec-configcheckannotations}
--- a/docs/configuration/crds/v1beta1/syslogng_types.md
+++ b/docs/configuration/crds/v1beta1/syslogng_types.md
@@ -25,7 +25,7 @@ SyslogNGSpec defines the desired state of SyslogNG
 
 ### configCheck (*ConfigCheck, optional) {#syslogngspec-configcheck}
 
-Overrides the default logging level configCheck setup. This field is not used directly, just copied over the field in the logging resource if defined. 
+Overrides the default logging level configCheck setup. This field is not used directly, its fields are copied over the ones in the logging resource, field by field, so setting one field here does not discard the others set on the logging resource. 
 
 
 ### configCheckPod (*typeoverride.PodSpec, optional) {#syslogngspec-configcheckpod}
--- a/pkg/sdk/logging/api/v1beta1/fluentd_types.go
+++ b/pkg/sdk/logging/api/v1beta1/fluentd_types.go
@@ -120,7 +120,7 @@ type FluentdSpec struct {
 	// Configure sidecar container in Fluentd pods, for example: [https://github.com/kube-logging/logging-operator/config/samples/logging_logging_fluentd_sidecars.yaml](https://github.com/kube-logging/logging-operator/config/samples/logging_logging_fluentd_sidecars.yaml).
 	SidecarContainers []corev1.Container `json:"sidecarContainers,omitempty"`
 	// Overrides the default logging level configCheck setup
-	// This field is not used directly, just copied over the field in the logging resource if defined
+	// This field is not used directly, its fields are copied over the ones in the logging resource, field by field, so setting one field here does not discard the others set on the logging resource
 	ConfigCheck *ConfigCheck `json:"configCheck,omitempty"`
 }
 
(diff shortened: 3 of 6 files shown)


vyncint:
The red `Go end2end tests` check on this PR is not caused by this change. It is a pre-existing flake in the e2e harness, which I have now traced and written up as #2287 with a fix in #2288.

Short version: of the last 100 `E2E tests` workflow runs, 7 failed, and all 7 are this same job, on branches with nothing in common — four `renovate/all` runs, `feat/support-raw-fluentd-filter`, `fix/hosttailer-namespace`, and this one. Five of them share an identical signature: the three `fluentbit-*` packages pass and the other ten all die with `panic: test timed out after 20m0s`.

The cause is that nothing bounds the `kind` calls the harness makes. In each bad run, 7–8 goroutines are blocked in `kind create cluster` and 5–6 in `kind load docker-image`. Counting progress lines, 17 clusters start "Preparing nodes" and only 9 finish — a full run creates exactly 17, so 8 never got a node container. `kind create cluster --wait 3m` looks like it should cover that, but `--wait` is the last action kind runs; node provisioning happens before it and is unbounded.

The artifact from this run is byte-for-byte the same shape as the known-unrelated `renovate/all` failure: only the three `fluentbit-*` cluster logs, because a package that panics never reaches its deferred log dump.

Two notes for whoever picks this up:

- This run's job has been wedged in `Post Set up Docker Buildx` for about six hours, so its logs are not downloadable through the API and the check will sit at `pending` until GitHub's 6-hour limit kills it. I diagnosed it from the uploaded artifact plus the identical failures on other branches. It will need a re-run.
- Nothing in this PR touches the e2e harness — it only changes `AggregatorLevelConfigCheck` and its tests.

vyncint:
Update: that job has now been killed at GitHub's 6-hour limit (`6h5m0s`), so the check has flipped from pending to red. Its logs were never uploaded — both the log and zip endpoints return 404 — so there is nothing to inspect even now; the diagnosis above stands on the uploaded artifact and the identical failures on unrelated branches. A re-run of `E2E tests` is all this needs.

csatib02 (approved):

Issue #2287, opened 2026-07-28
E2E: unbounded kind CLI calls turn a stalled cluster build into a 20-minute panic that takes the whole package down


vyncint:
**Describe the bug**:

The `Go end2end tests` job fails on roughly 7% of runs, on PRs that cannot possibly have caused it. When it goes wrong it usually takes ten of the thirteen packages down together, runs for an hour, and reports nothing more useful than `panic: test timed out after 20m0s`.

The cause is that nothing bounds the `kind` CLI calls the harness makes. When a cluster build stalls, the only thing that ends the stall is `go test` hitting its own `-timeout`, which panics the whole package binary — so the deferred cluster cleanup never runs and the leftover node containers make the next wave likelier to stall as well.

**Evidence**:

Of the last 100 `E2E tests` workflow runs, 7 failed. All 7 failures are this one job:

| run | branch | failure |
|---|---|---|
| [30310174413](https://github.com/kube-logging/logging-operator/actions/runs/30310174413) | `fix-configcheck-merge-2228` (#2286) | mass timeout |
| [30274939774](https://github.com/kube-logging/logging-operator/actions/runs/30274939774) | `renovate/all` | mass timeout |
| [30158701484](https://github.com/kube-logging/logging-operator/actions/runs/30158701484) | `renovate/all` | mass timeout |
| [30153096685](https://github.com/kube-logging/logging-operator/actions/runs/30153096685) | `renovate/all` | mass timeout |
| [29083924539](https://github.com/kube-logging/logging-operator/actions/runs/29083924539) | `feat/support-raw-fluentd-filter` | mass timeout |
| [29879121084](https://github.com/kube-logging/logging-operator/actions/runs/29879121084) | `fix/hosttailer-namespace` | single package |
| [29480219033](https://github.com/kube-logging/logging-operator/actions/runs/29480219033) | `renovate/all` | single package |

The five "mass timeout" runs are all the same shape: the three `fluentbit-*` packages pass, and the other ten fail with `panic: test timed out after 20m0s`.

In those runs the goroutine dumps are consistent to within one:

- **7–8** goroutines blocked in `common.KindClusterKubeconfig` → `kind create cluster`
- **5–6** goroutines blocked in `common.kindCluster.LoadImages` → `kind load docker-image`

So every hang is inside a `kind` subprocess that never returns. Counting the progress lines in one job:

```
17  • Preparing nodes 📦   ...
 9  ✓ Preparing nodes 📦
```

[...]

PR #2288, opened 2026-07-28
fix(e2e): bound kind CLI invocations with a timeout

vyncint:
Fixes part 1 of #2287.

Nothing bounds the `kind` calls the e2e harness makes. When a cluster build stalls, the only thing that ends it is `go test` hitting its own `-timeout` and panicking — which kills every other test in that package binary and skips the deferred cluster cleanup, so the leftover node containers make the next wave likelier to stall too. Ten of the thirteen packages going down together is the usual result.

This gives every `kind` invocation a deadline.

## What changes

`Kind` is the central object: it carries the binary path, the node image and both deadlines, and `CreateCluster`, `DeleteCluster`, `GetKubeconfig` and `LoadDockerImage` are methods on it. Its `run` helper wraps each invocation in `exec.CommandContext`. When the deadline is what stopped the process, the error wraps a new `ErrTimeout` and names the subcommand, because exec on its own only reports `signal: killed`:

```
kind create cluster --name fluentd-configcheck --wait 3m timed out after 16m0s (override with KIND_COMMAND_TIMEOUT)
```

`CreateCluster` also deletes the half-built cluster on timeout. kind cleans up after itself when one of *its* actions fails, but not when we kill it, so those containers would otherwise sit there holding their share of the runner. That delete gets its own shorter `CleanupTimeout`, since the runner is already struggling by then and the caller still has to report the failure inside the remaining test budget.

`cmd.WaitDelay` is set because kind shells out to docker and killing kind does not kill its children. Whenever output goes somewhere other than a file, exec pipes it and `Wait` blocks until every writer closes that pipe — so a surviving docker would keep us there well past the deadline. I measured this with a stub: 3.01s without the delay against a 100ms context, 352ms with it.

## Why `--wait` was never going to do this

`KindClusterCreationTimeout = "3m"` reads like it already bounds cluster creation. It does not. In kind v0.31.0, `p.Provision` — the "Preparing nodes" phase — runs *before* the action list, and `waitforready.NewAction(opts.WaitForReady)` is the last action in it. So `--wait` bounds the control plane readiness poll and nothing else, while the phase that actually stalls is unbounded. I have left the flag alone and put a comment on the const saying what it does and does not cover.

`kind load docker-image` has no `--wait` at all, and 5–6 of the stuck goroutines in each bad run are sitting in it.

## On the default

Derived, not chosen. `flag.Lookup("test.timeout")` reads the deadline the test binary was actually started with, and the cap is 80% of it — 16m under the suite's `-timeout 20m`, following `E2E_TEST_TIMEOUT` automatically if that changes. `KIND_COMMAND_TIMEOUT` still overrides it; a binary started with no deadline falls back to 15m.

An earlier revision of this PR hard-coded 10 minutes, picked by comparison with healthy timings. That was wrong, and it made the PR a regression: both e2e runs on the branch failed with the cap firing on `kind load docker-image`. Healthy timings are the wrong reference — the cap only matters on a degraded runner, which is exactly when every step legitimately takes several times longer. In the second run every package that passed took 1.6×–6.2× its healthy time, one of them 878s. Deriving from the enclosing budget means the cap can only convert an anonymous 20-minute panic into a named failure with cleanup, and cannot turn a passing run red.

## What this does and does not do

It does not stop the stalls. I proposed serialising the image loads on #2287 and then measured it: concurrency is 2.16× faster on an idle host, and serialising makes the tail ~3× worse under saturation — which is what a per-command deadline trips on, so it would have caused *more* timeouts, not fewer. Retracted there. What the numbers point at instead is loading *less*: `setup.LoggingOperator` puts all six images into every cluster regardless of the test, ~779 MB × 17 ≈ 13 GB per run, `fluentd-full` alone 67% of it. Left on the issue, as it is too big to bolt on here.

What it does: the failure arrives at 16 minutes instead of 20, says which command stalled instead of `panic: test timed out`, does not take the other tests in the package down with it, and does not leak node containers into the next wave.

## Testing

**Unit tests**, `e2e/common/kind/commands_test.go`, 10 tests / 36 counting subtests, table-driven. Each case builds its own `Kind` pointed at a stub binary, so the stall path is exercised without building anything; the suite runs in ~3.2s. They run under `make test` rather than the e2e job — `E2E_TEST` has no default, so the e2e recipe expanded to `./...` and swept them in; the Makefile change moves them to the unit target and filters `common` out of the e2e package list. The stub is POSIX-only — I ran the suite with its shebang switched to `dash`, since that is `/bin/sh` on the runners and not what I have locally.

Negative controls, to check the tests actually catch this:

| mutation | result |
|---|---|
| deadline removed (back to `exec.Command`) | `panic: test timed out after 40s` — the production symptom, in miniature |
| partial-cluster cleanup removed | 2 failures, incl. `[]string{"create cluster --name stuck"} does not contain "delete cluster --name stuck"` |

There is one thing I tried and dropped: a test that an orphaned grandchild holding the output pipe cannot outlast the deadline. Whether the shell leaves such an orphan turns out to depend on its fork/exec optimisation — I got 3.01s and 101ms from scripts differing by one line — so the test would have been a coin flip on someone else's runner. `WaitDelay` stays, measured but not asserted on.

**Against real kind** (v0.29.0, Docker 29.6.2, arm64), driving the changed functions directly:

```
create ok in 28s
kubeconfig ok, 5659 bytes
load ok in 1s
delete ok in 1s
```

and the case this PR is about, with the deadline set below a real cluster build:

```
create stopped after 20s with: kind create cluster --name timeout-verify-stall --wait 3m
  timed out after 20s (override with KIND_COMMAND_TIMEOUT)
no "timeout-verify-stall" cluster left behind
```

A full `make test-e2e E2E_TEST=watch-selector` is building locally as I write this, to confirm the normal `WithCluster` path is unaffected; I will report the result below. The e2e job on this PR covers the same ground more thoroughly.

While verifying I hit a `kind load` failure on `busybox:latest` (`content digest ... not found`) that turned out to be a local arm64 multi-platform quirk — `kind load docker-image busybox:latest` fails identically with none of this code involved. Worth mentioning only because the new code surfaced kind's real error in full rather than mislabelling it a timeout, which is the passthrough path working.

`golangci-lint` is clean on `e2e/common/kind`, and `make lint` still reports 0 issues. Note that `make lint` does not cover the `e2e` module — it runs at the repo root and in `pkg/sdk` only — so I ran it there by hand. It has 7 pre-existing findings in files I have not touched; I left those alone. I compared against a worktree of unmodified HEAD: the same 7 before and after, none introduced here.

### DIFF
--- a/e2e/common/kind/commands.go
+++ b/e2e/common/kind/commands.go
@@ -28,91 +28,179 @@ import (
 	"time"
 )
 
-var (
-	KindPath  string
-	KindImage string
-)
-
-// ErrTimeout is returned when a kind invocation is stopped by its deadline
-// rather than by kind itself.
+// ErrTimeout reports that an invocation was stopped by its deadline rather than
+// by kind itself.
 var ErrTimeout = errors.New("timed out")
 
-// CommandTimeout bounds a single kind invocation. Zero means it is derived
-// from the test binary's own -timeout on first use; see commandTimeout.
-//
-// kind's own --wait only covers the last of its create actions, waiting for
-// control plane readiness. Provisioning the node containers, pulling the node
-// image and `kind load docker-image` are bounded by nothing. When several
-// clusters are built at once on a busy CI runner those phases can stall for
-// longer than the whole test budget, and with no deadline here the only thing
-// that ends the stall is `go test` panicking on its own -timeout. That kills
-// every other test in the package and skips the deferred cluster cleanup, so
-// one stalled cluster takes the rest of the package down with it.
-//
-// Set it directly, or with KIND_COMMAND_TIMEOUT, to pin a value.
-var CommandTimeout time.Duration
+const (
+	pathEnv           = "KIND_PATH"
+	imageEnv          = "KIND_IMAGE"
+	commandTimeoutEnv = "KIND_COMMAND_TIMEOUT"
 
-// timeoutFraction is how much of the enclosing -timeout a single kind call may
-// consume before we call it stalled.
-//
-// The deadline has to be derived from the budget it sits inside, not guessed
-// from how long a healthy run takes. Healthy timings are the wrong reference:
-// this only fires when the runner is degraded, and that is exactly when every
-// step legitimately takes several times longer than healthy. A cap chosen from
-// healthy numbers ends up inside the range where real work still completes,
-// and then it fails runs that would have passed.
-//
-// Anything still running this late has left too little of the budget for the
-// rest of the test to finish, so cutting it can only replace an anonymous
-// whole-binary panic with a named failure and a cleaned-up cluster. It cannot
-// turn a passing run red.
-const timeoutFraction = 0.8
+	defaultPath = "../../bin/kind"
 
-// fallbackTimeout applies when the enclosing deadline cannot be read, which
-// means the binary was started with -timeout 0 and has no deadline of its own.
-const fallbackTimeout = 15 * time.Minute
+	// timeoutFraction is the share of the enclosing -timeout that one invocation
+	// may consume. The cap is derived from the surrounding budget rather than
+	// from healthy timings because a stall only matters on a degraded runner,
+	// which is exactly when every step legitimately takes several times longer.
+	timeoutFraction = 0.8
 
-// CleanupTimeout bounds the best-effort delete issued after a command times
-// out. It is much shorter than CommandTimeout: the runner is already
-// struggling by then, and the caller still has to report the failure inside
-// the remaining test budget.
-var CleanupTimeout = 2 * time.Minute
+	// fallbackTimeout applies when the binary has no deadline to derive from.
+	fallbackTimeout = 15 * time.Minute
 
-const commandTimeoutEnv = "KIND_COMMAND_TIMEOUT"
+	defaultCleanupTimeout = 2 * time.Minute
+	defaultWaitDelay      = 10 * time.Second
+)
 
-// waitDelay is how long Wait may keep waiting on inherited output pipes after
-// the process itself has been killed. A variable so the tests can shorten it.
-var waitDelay = 10 * time.Second
+// Kind runs kind subcommands against one installation of the binary.
+type Kind struct {
+	Path  string
+	Image string
 
-func init() {
-	KindPath = os.Getenv("KIND_PATH")
-	if KindPath == "" {
-		KindPath = "../../bin/kind"
-		fmt.Println("KIND_PATH is not set, defaulting to ../../bin/kind")
+	// CommandTimeout bounds a single invocation. Zero derives it from the test
+	// binary's own -timeout on first use.
+	CommandTimeout time.Duration
+
+	// CleanupTimeout bounds the delete issued after a timeout. It is short
+	// because the runner is already struggling by then and the caller still has
+	// to report the failure within what is left of the budget.
+	CleanupTimeout time.Duration
+
+	// waitDelay caps how long Wait blocks on inherited pipes after the process
+	// has been killed.
+	waitDelay   time.Duration
+	resolveOnce sync.Once
+}
+
+// New returns a Kind configured from KIND_PATH, KIND_IMAGE and
+// KIND_COMMAND_TIMEOUT.
+func New() *Kind {
+	path := os.Getenv(pathEnv)
+	if path == "" {
+		path = defaultPath
+		fmt.Printf("%s is not set, defaulting to %s\n", pathEnv, defaultPath)
 	}
-	KindImage = os.Getenv("KIND_IMAGE")
-	// Not resolved here: the enclosing -timeout is only readable once the
-	// testing package has parsed flags, which happens after init.
-	CommandTimeout = resolveCommandTimeout(os.Getenv(commandTimeoutEnv), 0)
+
+	return &Kind{
+		Path:  path,
+		Image: os.Getenv(imageEnv),
+		// Left to be derived: the enclosing -timeout only becomes readable once
+		// the testing package has parsed flags, which is after initialisation.
+		CommandTimeout: resolveCommandTimeout(os.Getenv(commandTimeoutEnv), 0),
+		CleanupTimeout: defaultCleanupTimeout,
+		waitDelay:      defaultWaitDelay,
+	}
 }
 
-// commandTimeout is the deadline for one kind call, worked out once.
-func commandTimeout() time.Duration {
-	resolveOnce.Do(func() {
-		if CommandTimeout > 0 {
-			return // pinned by KIND_COMMAND_TIMEOUT or by a caller
-		}
-		CommandTimeout = deriveCommandTimeout(enclosingTestTimeout())
-		fmt.Printf("kind command timeout: %s (override with %s)\n", CommandTimeout, commandTimeoutEnv)
+func (k *Kind) CreateCluster(options CreateClusterOptions) error {
+	if k.Image != "" && options.Image == "" {
+		options.Image = k.Image
+	}
+	args := options.AppendToArgs([]string{"create", "cluster"})
+
+	cmderr := &bytes.Buffer{}
+	err := k.run(k.commandTimeout(), args, func(cmd *exec.Cmd) {
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = io.MultiWriter(os.Stderr, cmderr)
 	})
-	return CommandTimeout
+
+	switch {
+	case err == nil:
+		return nil
+
+	case errors.Is(err, ErrTimeout):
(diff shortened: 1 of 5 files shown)


vyncint:
Local full-package run finished, as promised above:

```
make test-e2e E2E_TEST=watch-selector
--- PASS: TestWatchSelectors (116.43s)
ok  github.com/kube-logging/logging-operator/e2e/watch-selector  117.451s
EXIT=0
```

That is the whole `WithCluster` path — create, load images, install the operator, reconcile, tear down — through the changed functions, and 116s against the 131s this package takes on a green CI run, so the deadline adds nothing measurable. `kind get clusters` is empty afterwards, and `codegen`/`manifests` left the tree clean.

One more compatibility note: the test stub is POSIX-only, and I ran the unit suite with its shebang switched to `dash` as well as `sh`, since `/bin/sh` on the runners is dash and is not what I have locally. Green both ways.

e2e/common/kind/commands.go
```diff
@@ -34,6 +69,49 @@ func init() {
...
+func resolveCommandTimeout(raw string, fallback time.Duration) time.Duration {
+	if raw == "" {
+		return fallback
+	}
+	timeout, err := time.ParseDuration(raw)
+	if err != nil {
+		fmt.Printf("%s=%q is not a valid duration, keeping %s\n", commandTimeoutEnv, raw, fallback)
+		return fallback
+	}
+	return timeout
```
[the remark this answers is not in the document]
vyncint: Good catch, fixed in `7db6906`.

`0` was the dangerous one: `time.ParseDuration("0")` succeeds, and a zero deadline is already expired, so `KIND_COMMAND_TIMEOUT=0` would have failed every kind call instantly. Anyone writing `0` almost certainly means "no timeout" — exactly the opposite of what they would have got. Negative values behaved the same way.

Non-positive now keeps the default and says so. I chose that over honoring `0` as unbounded, since restoring the unbounded behaviour is the thing this PR exists to remove; a large duration is the escape hatch if anyone wants one.

Covered by three new cases in `TestResolveCommandTimeout` (`0`, `0s`, `-1s`); removing the guard fails all three.

e2e/common/kind/commands.go
```diff
@@ -42,14 +120,28 @@ func CreateCluster(options CreateClusterOptions) error {
...
+		// but not when we kill it, so the leftovers have to go explicitly.
+		// Otherwise the node containers hold their share of the runner for the
+		// rest of the job and make the next cluster likelier to stall too.
+		if deleteErr := deleteCluster(CleanupTimeout, DeleteClusterOptions{Name: options.Name}); deleteErr != nil {
+			return fmt.Errorf("%w; deleting the partial cluster also failed: %w", err, deleteErr)
+		}
+		return err
+	default:
 		return errors.New(cmderr.String())
 	}
```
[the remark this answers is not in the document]
vyncint: Agreed, fixed in `7db6906`. This one predates the PR, but it is in the branch I restructured, so it belongs here.

`CreateCluster` now falls back to the exec error when stderr is empty, and still prefers kind's stderr whenever there is any — that matters, because `common.isClusterAlreadyExistsError` matches against that exact text, so always wrapping would have been the riskier change.

Two tests, one for each side: `TestCreateClusterReportsTheExecErrorWhenKindWritesNothing` points `KindPath` at a missing binary, and `TestCreateClusterPrefersKindsStderrWhenThereIsAny` pins the already-exists text. Dropping the fallback reproduces exactly what you described:

```
Error:    Should NOT be empty, but was
Messages: the error must say something
```

with the stderr test still passing, so the two are independent.

vyncint:
**Please don't merge this yet — the second e2e run convinced me the default in here is wrong, and I think this PR is currently a net regression.**

Both e2e runs on this branch failed, and in both the failures are my own timeout firing on the same operation:

```
run 1 (6320489b): 6 hits, all `kind load docker-image ... fluentd-full:local`, 10m0s
run 2 (7db69062): 5 hits, all `kind load docker-image ... fluentd-full:local`, 10m0s
```

Two failures out of two, against roughly a 7% base rate on other branches, with every other e2e run in the last day green. I can't call that the pre-existing flake.

**Where my reasoning went wrong.** I picked 10 minutes by comparing it to healthy timings: the slowest whole package on a green run is 581s, so 600s for one call looked generous. That compares the wrong things. The timeout only matters when the runner is degraded, and degraded is exactly when everything takes several times longer. Run 2 makes the point — every package that passed was 1.6× to 6.2× its healthy time:

| package | healthy | run 2 | |
|---|---|---|---|
| `fluentd-aggregator-detached-multiple-failures` | 143s | **878s** | 6.2× |
| `fluentd-aggregator-namespacelabel` | 127s | 441s | 3.5× |
| `volumedrain` | 207s | 528s | 2.6× |
| `watch-selector` | 132s | 328s | 2.5× |
| `syslog-ng-aggregator-detached` | 187s | 427s | 2.3× |

Eight packages passed at those speeds. `fluentd-aggregator-detached-multiple-failures` loaded the same 522 MB `fluentd-full` image and got through in 878s. So under this contention the loads do complete — they are slow, not wedged — and a 10-minute cap cuts off work that would have finished. That is the opposite of what I intended, and it is my bug, not the flake.

**What I think the right shape is.** The cap should be derived from the deadline it sits inside, not from healthy-case timings. `go test` runs with `-timeout 20m`; anything still running near that was going to take the package down regardless, so a cap just under it can only ever convert an anonymous 20-minute panic into a named failure with cleanup, and can never turn a passing run red. Ten minutes leaves 10 minutes on the table and is well inside the range where real work still completes.

That alone is a small win though. The thing that would actually fix this is the image-load serialisation from #2287: five clusters unpacking the same 522 MB image into containerd simultaneously is what makes a 90-second load take more than ten minutes. With loads serialised, the timeout goes back to being a backstop that never fires.

So I'd rather not land this as-is. I'll push a corrected default and add the load mutex, unless you'd prefer the two split across separate PRs — happy either way, just say which.

Sorry for the noise; better to catch it here than after it merged.

vyncint:
Pushed `164b5da`, and I measured the serialisation idea before proposing it further — it turned out to be wrong, so I have retracted it on #2287.

**The timeout is now derived, not guessed.** `flag.Lookup("test.timeout")` reads the deadline the test binary was actually started with, and the cap is 80% of it — 16m under the suite's `-timeout 20m`, and it follows automatically if `E2E_TEST_TIMEOUT` ever changes. `TestDerivedTimeoutStaysBelowTheEnclosingDeadline` pins the invariant; setting the fraction to 1.0 fails it for every budget:

```
Error: "20m0s" is not less than "20m0s"
```

**On the load mutex I floated in #2287 — I measured it and it is a bad idea.** Loading the 522 MB fluentd image into N fresh clusters:

| | idle host | saturated host, last load done |
|---|---|---|
| serial | 23.3s (4 clusters) | **290.4s** (6 clusters) |
| concurrent | **10.8s** | **99.2s** |

Concurrency is 2.16× faster idle and holds up under saturation. Serialising would make the tail ~3× worse — and the tail is exactly what a per-command deadline trips on. So that would have increased timeouts, not reduced them. Good thing it never got written.

What the numbers do point at is loading *less*: `setup.LoggingOperator` puts all six images into every cluster regardless of the test, ~779 MB × 17 clusters ≈ 13 GB per run, and `fluentd-full` alone is 67% of that. Details on the issue. That is too big to bolt onto this PR, so I have left it there.

25 unit tests green, lint clean. Marking ready for review again — this change should now be strictly non-regressive: it can convert a 20-minute anonymous panic into a named failure with cleanup, and cannot cause one.

e2e/common/kind/commands_test.go
csatib02, line 1: We prefer table-driven tests, where applicable!

e2e/common/kind/commands_test.go
```diff
@@ -0,0 +1,321 @@
+// Copyright © 2021 Cisco Systems, Inc. and/or its affiliates
```
csatib02: The current license header is the kube-logging authors format, the date is also invalid.

e2e/common/kind/commands.go
csatib02, line 1: I think we can re-design the current approach and introduce a more structured approach, with a kind object being a central component, and commands deriving from it.

Also, please make sure your comments are kept concise, despite doing development with AI, it's really important that the code speaks for itself, and comments are placed on hard to understand parts.

e2e/common/kind/commands_test.go
csatib02, line 1: These tests, run together with the e2e-tests, but they should rather run with the unit-tests.

csatib02:
Thanks @vyncint for handling this flakiness!
I'm going to merge this PR, but the e2e test-suite needs some caring, and I have an idea how to restructure it.

csatib02 (approved):

Issue #2291, opened 2026-07-29
E2E: the harness shape is generating the bugs — restructure e2e/ around a lifecycle-owning harness


vyncint:
**Is your feature request related to a problem? Please describe.**

The `e2e` module has grown to the point where the per-test boilerplate is larger than the tests, and that boilerplate is where the bugs are. Everything below was measured on `master` @ `a252292c`.

The module is 7004 lines of Go across 13 suite directories, with 2096 of those in `e2e/common/`. There are 17 test functions, and every one of them provisions its own KIND cluster (11 suites × 1, `volumedrain` × 2, `fluentd-aggregator` × 4).

The entry point every suite goes through is:

```go
func WithCluster(name string, t *testing.T, fn func(*testing.T, Cluster), beforeCleanup func(*testing.T, Cluster) error, opts ...cluster.Option)
```

— `common/cluster.go:52`. Five positional parameters, two of them differently-shaped anonymous closures. Because setup and teardown are both callback bodies rather than anything reusable, each of the 17 call sites is a self-contained nested block: **median 206 lines, longest 699** (`TestElasticsearch_MultiVersion`), shortest 104.

What that shape has produced, counted across the module:

| | count |
|---|---|
| `var TestTempDir` + `init()` reading `PROJECT_DIR`, byte-identical | 13 (one renamed `TestTempDirUnnamed`) |
| scheme-registration `AddToScheme` lines | 103 |
| `beforeCleanup` PrintLogs/coverage closures | 17, in 3 divergent variants |
| references to the `common.<Image>Repo` / `Tag` constants | 186 |
| `5*time.Minute, 3*time.Second` literal pairs | 17 |
| `context.Background()` | 26 |
| `t.Context()`, though the module is `go 1.26.0` | 0 |
| `kubectl` shell-outs | 22 |
| `common/panicobject.go`, zero references outside itself | 133 lines |

Two structural causes, both of which block incremental cleanup:

1. **Inverted package containment.** `common/cond` and `common/setup` are subpackages of `common` yet import it — `cond/conditions.go:22`, `setup/loggingoperator.go:29`, `setup/logproducer.go:30`. Anything the three layers share therefore has to live in `common`, which is why `common` is a grab-bag and why it cannot be split in place.
2. **The module is not linted.** `make lint` covers the repo root and `pkg/sdk`; `e2e` is a separate module and is skipped. Under the pinned `golangci-lint 2.12.2` it has **45 findings**:

   | linter | count |
[...]

PR #2292, opened 2026-07-29
build(e2e): run the linter on the e2e module, and fix what it reports

vyncint:
Part of #2291, and independent of it — this stands on its own whether or not the restructure proposed there goes ahead.

`e2e` is a separate Go module, so the existing `make lint` — which covers the repo root and `pkg/sdk` — has never linted it. This adds it, and fixes everything that surfaces.

Enabling and fixing have to be one change: `make lint` is what CI runs at `.github/workflows/ci.yaml:53`, so turning the module on without clearing the backlog would just make CI red.

### The one finding that is not style

`e2e/common/setup/loggingoperator.go:96` assigned the error from `actionConfig.Init(...)` and never checked it:

```go
err = actionConfig.Init(restClientGetter, opt.Namespace, "memory", func(...) {...})

installer := action.NewInstall(actionConfig)          // err never checked
...
cp, err := installer.LocateChart(...)                 // err overwritten here
```

`err` is reassigned fifteen lines later by `LocateChart`, so a failed helm action-config init was silently discarded and every e2e run proceeded to install the chart through a half-initialized `action.Configuration`. `ineffassign` is the only reason this was visible at all, which is a fair argument for the rest of the change. It gets its own commit.

### How many findings there actually are

45, on `master` @ `a252292c`:

| linter | count |
|---|---|
| `gofumpt` | 18 |
| `whitespace` | 16 |
| `gci` | 4 |
| `staticcheck` | 3 |
| `goimports` | 2 |
| `errcheck` | 1 |
| `ineffassign` | 1 |

Worth knowing before anyone tries to reproduce that: **`golangci-lint run` prints only 12 of the 45 by default.** `max-same-issues` defaults to 3 and applies per unique message text; `gci`, `gofumpt` and `goimports` all emit the identical string "File is not properly formatted", so 24 findings collapse into 3 printed lines, and `whitespace`'s 16 collapse into 4. Fixing the visible 12 only uncovers the next batch — I did that first and had to go round again. The full list needs:

```bash
cd e2e && golangci-lint run --max-same-issues=0 --max-issues-per-linter=0
```

This also means issue #2287's "currently has 8 findings" was wrong twice over; I've corrected it there.

### Commits

| | |
|---|---|
| `fix(e2e): check the error returned by the helm action config init` | the bug above |
| `fix(e2e): report the log file close error from PrintLogs` | `errcheck` — the deferred `Close` on stern's output file dropped its error, so a truncated cluster log dump looked like a successful one |
| `refactor(e2e): precompile the config check pattern, drop a redundant selector` | `staticcheck` — `regexp.MatchString` recompiled the same pattern on every iteration of two `Eventually` loops (SA6000); `ChartPathOptions` is embedded in `action.Install`, so naming it was redundant (QF1008) |
| `chore(e2e): format the module with gofumpt, gci and goimports` | the 40 formatter and whitespace findings |
| `build(e2e): run the linter on the e2e module` | the `Makefile` change, in `lint` and `lint-fix` both |

The formatting commit is mechanical and carries no behaviour change: octal literals gain the `0o` prefix, `github.com/kube-logging/...` imports move into their own `gci` group, and `gofumpt` adds the blank lines and trailing commas it wants. I ran the formatter against the specific files the linter flagged rather than across the module, so nothing outside the reported set is touched — 18 files, all of them ones with findings.

### Verification

`make check` (`license-check lint test`) passes.

No new findings introduced — unmodified `master` checked out in a separate worktree, same command both sides:

```
BEFORE (master @ a252292c): 45 findings
AFTER  (this branch):        0 findings
findings present after but not before: none
```

`make lint` now covers all three modules and is clean:

```
$ make lint
bin/golangci-lint run --timeout 10m
0 issues.
cd pkg/sdk && bin/golangci-lint run --timeout 10m
0 issues.
cd e2e && bin/golangci-lint run --timeout 10m
0 issues.
```

Measured with the pinned linter, `golangci-lint 2.12.2` (`Makefile:11`), on Go 1.26.5.

### Scope

Lint enablement and the findings it reports, nothing else. No test behaviour changes, no restructuring. The `e2e` module has other problems — `t.FailNow()` called from a non-test goroutine at 15 sites, three `30*time.Minute` waits sitting under a 20-minute package timeout, and an inconsistent coverage-collection policy — but those are separate fixes and I have written them up in #2291 rather than folding them in here.

### DIFF
--- a/e2e/common/cluster.go
+++ b/e2e/common/cluster.go
@@ -156,16 +156,12 @@ func (c kindCluster) PrintLogs(config PrintLogConfig) error {
 	if err != nil {
 		return err
 	}
+	defer func() { _ = f.Close() }()
+
 	cmd.Stdout = f
 	cmd.Stderr = os.Stderr
 
-	// f is the log file stern writes into, so a failed Close means truncated output.
-	if err := cmd.Run(); err != nil {
-		_ = f.Close()
-		return err
-	}
-
-	return f.Close()
+	return cmd.Run()
 }
 
 func (c kindCluster) Cleanup() error {
--- a/e2e/common/cluster.go
+++ b/e2e/common/cluster.go
@@ -156,12 +156,16 @@ func (c kindCluster) PrintLogs(config PrintLogConfig) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-
 	cmd.Stdout = f
 	cmd.Stderr = os.Stderr
 
-	return cmd.Run()
+	// f is the log file stern writes into, so a failed Close means truncated output.
+	if err := cmd.Run(); err != nil {
+		_ = f.Close()
+		return err
+	}
+
+	return f.Close()
 }
 
 func (c kindCluster) Cleanup() error {
--- a/e2e/common/cluster.go
+++ b/e2e/common/cluster.go
@@ -92,7 +92,7 @@ func GetTestCluster(clusterName string, opts ...cluster.Option) (Cluster, error)
 	if err != nil {
 		return nil, errors.WrapIfWithDetails(err, "unable to create temp file for kubeconfig", "clusterName", clusterName)
 	}
-	err = os.WriteFile(kubeconfigFile.Name(), kubeconfig, os.FileMode(0600))
+	err = os.WriteFile(kubeconfigFile.Name(), kubeconfig, os.FileMode(0o600))
 	if err != nil {
 		return nil, errors.WrapIfWithDetails(err, "failed to write kubeconfig", "clusterName", clusterName, "path", kubeconfigFile.Name())
 	}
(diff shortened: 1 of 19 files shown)


e2e/common/cluster.go
```diff
@@ -156,12 +156,16 @@ func (c kindCluster) PrintLogs(config PrintLogConfig) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
```
csatib02, line 159: I dont really see why this change makes any difference, besides having cmd.Run's err captured.

e2e/common/setup/loggingoperator.go
```diff
@@ -96,6 +93,9 @@ func LoggingOperator(t *testing.T, c common.Cluster, opts ...LoggingOperatorOpti
 	err = actionConfig.Init(restClientGetter, opt.Namespace, "memory", func(format string, v ...interface{}) {
 		t.Logf(format, v...)
 	})
+	if err != nil {
+		t.Fatalf("helm action config init: %s", err)
+	}
```
csatib02: Lets block scope this.

csatib02 (approved):

csatib02 (approved):

PR #2298, opened 2026-07-29
refactor(e2e): move common/cond to internal/wait and take plain values

vyncint:
§5.2 of #2291, following #2297. Moves the conditions package to `internal/wait` and fixes two things about it on the way.

### The signatures took pointers to an interface and to a context

Four helpers looked like this:

```go
func CheckFluentdStatus(t *testing.T, c *common.Cluster, ctx *context.Context, ...) bool {
	cluster := *c
```

`common.Cluster` is an interface, so `*common.Cluster` is a pointer to an interface, dereferenced immediately. The `context.Context` pointer was dereferenced at every use. Neither pointer bought anything.

They only ever called `cluster.GetClient()`, so they now take a `client.Client` directly — which is what the five conditions sitting beside them in the same file already took. Nine call sites updated.

### They failed the test from testify's goroutine

Each of those helpers refreshed its object through `common.RequireNoError`, which calls `t.FailNow()`. All of them are invoked from inside `require.Eventually`, and testify runs the condition on a goroutine of its own — v1.11.1, `assert/assertions.go`:

```go
checkCond := func() { ch <- condition() }
...
go checkCond()
```

So that was `t.FailNow()` from a non-test goroutine, which the `testing` docs say must not happen — the same defect as #2295, in a second place. Twelve calls.

A transient read failure inside a polling condition should not end the run anyway; it should be retried. The refresh now logs and returns, and the condition returns false as it already did, so the next tick picks it up:

```go
func refresh(t *testing.T, cl client.Client, ctx context.Context, obj client.Object) {
	if err := cl.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
		t.Logf("could not refresh %s: %v", obj.GetName(), err)
	}
}
```

**This is not the whole problem.** Twenty more `RequireNoError` calls sit directly inside `Eventually` conditions in three suites — `fluentd-aggregator-detached`, `syslog-ng-aggregator-detached` and `fluentd-aggregator`. Those are suite-local and I have left them alone rather than widen this PR; happy to follow up.

### The package no longer imports `common`

Those were the only two things it used from `common` — the `Cluster` interface and `RequireNoError`. With both gone, `internal/wait` imports nothing from `common`. #2291 describes that upward dependency as the thing preventing `common` from being split, so this is one package out from under it.

### Tests

Six table-driven tests, sixteen cases, over `fake.NewClientBuilder()` — no cluster, 0.07 s. These and `internal/kind` are the module's only unit tests.

They are picked up automatically by the `./internal/...` selector added in #2297, so `make check` now reports 28 packages rather than 27.

Checked they actually catch something: replacing the ready-replica comparison with `if false` fails exactly `TestDeploymentAvailable/not_all_replicas_ready`, and restoring it passes.

### What is deferred

§5.2 also calls for named wait budgets to replace the seventeen `5*time.Minute, 3*time.Second` literals. That is a change to timeout behaviour across suites, and the divergent values (`syslog-ng-*` poll at 2 s, `volumedrain` uses eight distinct pairs) need measuring before anything is homogenised. It belongs in its own PR.

Function names are unchanged. `wait.CheckFluentdStatus` reads oddly, but renaming sixty-seven call sites for style, in the same PR that moves them, would bury the two real fixes.

### Verification

`make check` passes. `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` in `e2e/` reports 0 issues. Both commits build on their own.

### Scope

14 files. The package itself, plus an import path and a `cond.` → `wait.` qualifier in each of the twelve suites that use it.

### DIFF
--- a/e2e/internal/wait/conditions.go
+++ b/e2e/internal/wait/conditions.go
@@ -16,12 +16,12 @@ package wait
 
 import (
 	"context"
-	"strconv"
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
@@ -102,22 +102,6 @@ func ResourceShouldBePresent(t *testing.T, cl client.Reader, obj client.Object) 
 	}
 }
 
-// activeIs reports whether Status.Active has been written and equals want. The
-// field is optional, so a nil pointer means the controller has not set it yet:
-// neither true nor false, and the caller should keep polling. Dereferencing it
-// blindly would panic on testify's condition goroutine and take the whole
-// package binary down with it.
-func activeIs(v *bool, want bool) bool { return v != nil && *v == want }
-
-// activeState renders Status.Active for logs, keeping "not written yet"
-// distinguishable from an explicit false.
-func activeState(v *bool) string {
-	if v == nil {
-		return "unset"
-	}
-	return strconv.FormatBool(*v)
-}
-
 // refresh re-reads obj so the next poll sees current status. A failed read is
 // logged and ignored on purpose: the caller is about to return false and be
 // retried, and these conditions run on testify's own goroutine, where failing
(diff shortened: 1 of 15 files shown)


csatib02 (approved):

e2e/internal/wait/conditions.go
```diff
@@ -101,6 +102,22 @@ func ResourceShouldBePresent(t *testing.T, cl client.Reader, obj client.Object)
...
+func activeIs(v *bool, want bool) bool { return v != nil && *v == want }
+
+// activeState renders Status.Active for logs, keeping "not written yet"
+// distinguishable from an explicit false.
+func activeState(v *bool) string {
+	if v == nil {
+		return "unset"
+	}
+	return strconv.FormatBool(*v)
+}
```
csatib02: What do we gain from this exactly?

vyncint: Status.Active is *bool with omitempty, so it is nil until the controller writes it, with Problems set and Active unset the old deref panics, and testify runs conditions on go checkCond(), so that kills the whole package binary rather than one test. activeIs also makes unset mean "keep polling" instead of letting CheckExcess* pass on an unreconciled object. activeState only keeps the log reading unset rather than a pointer address, happy to drop it.

csatib02: There is `apimeta.IsStatusConditionTrue()` which we can use here, also I think we can push this to the caller side with `assert.Equal()` and `assert.True()` or smth.

csatib02 (approved):

PR #2299, opened 2026-07-29
refactor(e2e): add internal/fixture with the CR and tenancy builders

vyncint:
Adds `e2e/internal/fixture`, builders for the CR literals the suites repeat. Nothing imports it yet, so no existing behaviour changes.

- `Logging`, `FluentdSpec`, `Buffer`, `HTTPOutput` and `Flow`, with functional options.
- `LoggingInfra`, `LoggingTenant` and `LoggingRoute` ported from `common/helpers.go` as pure builders returning the objects in creation order, so the caller creates them and the package needs no client, context or `testing.T`.
- Each default is the value most call sites already use. Anything that diverges passes an option.
- Tenancy keeps its own aggregator spec: it disables the PVC and requests less than the single-tenant suites.
- The package does not import `common`. The image constants are duplicated while both exist and a test pins them equal, so they cannot drift the way the syslog-ng reloader name did.

13 tests, no cluster. `make check` passes.

### DIFF
--- a/e2e/internal/fixture/fixture_test.go
+++ b/e2e/internal/fixture/fixture_test.go
@@ -95,8 +95,7 @@ func TestBufferDefaultsAndOverrides(t *testing.T) {
 	b := Buffer("time")
 	require.Equal(t, "file", b.Type)
 	require.Equal(t, "time", *b.Tags)
-	// Literals, not the constants: a constant asserted against itself proves
-	// nothing, and neither value is pinned by the literal test below.
+	// Literals, not the constants: a constant asserted against itself proves nothing.
 	require.Equal(t, "1s", b.Timekey)
 	require.Equal(t, "0s", b.TimekeyWait)
 
@@ -125,9 +124,8 @@ func TestHTTPOutputAndFlow(t *testing.T) {
 func TestBuilderReproducesTheMultiWorkerLiteral(t *testing.T) {
 	const ns = "testing-1"
 
-	// The Logging fluentd-aggregator's MultiWorker test writes by hand, minus the
-	// ObjectMeta.Namespace it sets: the API server clears that field for a
-	// cluster-scoped kind, so the builder does not set it.
+	// The literal fluentd-aggregator writes by hand, minus the ObjectMeta.Namespace
+	// the API server clears for a cluster-scoped kind.
 	want := &v1beta1.Logging{
 		ObjectMeta: metav1.ObjectMeta{Name: "fluentd-aggregator-multiworker-test"},
 		Spec: v1beta1.LoggingSpec{
--- a/e2e/internal/fixture/fixture_test.go
+++ b/e2e/internal/fixture/fixture_test.go
@@ -49,8 +49,8 @@ func TestLoggingDefaults(t *testing.T) {
 	l := Logging("ns", "lg")
 
 	require.Equal(t, "lg", l.Name)
-	require.Equal(t, "ns", l.Namespace)
 	require.Equal(t, "ns", l.Spec.ControlNamespace)
+	require.Empty(t, l.Namespace, "Logging is cluster-scoped")
 	require.True(t, l.Spec.EnableRecreateWorkloadOnImmutableFieldChange)
 
 	// Neither half is present until asked for: suites that want only an agent
@@ -138,11 +138,11 @@ func TestHTTPOutputAndFlow(t *testing.T) {
 func TestBuilderReproducesTheMultiWorkerLiteral(t *testing.T) {
 	const ns = "testing-1"
 
+	// The Logging fluentd-aggregator's MultiWorker test writes by hand, minus the
+	// ObjectMeta.Namespace it sets: the API server clears that field for a
+	// cluster-scoped kind, so the builder does not set it.
 	want := &v1beta1.Logging{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "fluentd-aggregator-multiworker-test",
-			Namespace: ns,
-		},
+		ObjectMeta: metav1.ObjectMeta{Name: "fluentd-aggregator-multiworker-test"},
 		Spec: v1beta1.LoggingSpec{
 			EnableRecreateWorkloadOnImmutableFieldChange: true,
 			ControlNamespace: ns,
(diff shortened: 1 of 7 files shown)


vyncint:
Both suppressed comments on the last review were right. Fixed in 96b3131.

`LoggingTenant` took `nsTenant` but set `WatchNamespaces` to `TenantRef`, so the aggregator watched a namespace named after the loggingRef rather than the one its `Flow` and `Output` are created in. It came over that way from `common/helpers.go:231`, which hard-codes the literal `"tenant"`. Both call sites — `fluentbit-multitenant` and `fluentbit-hotreload` — pass `nsTenant := "tenant"`, so the two strings coincide there and the bug never fires. Nothing in `common` needs changing, and since the suites pass `"tenant"`, the objects the builder produces for them are unchanged by this.

The second comment was the more useful half: the test passed `"tenant-ns"` and asserted `[]string{TenantRef}`, so it pinned the trap rather than catching it. It now asserts against the `Flow`'s own namespace, plus a guard that the fixture namespace differs from `TenantRef` — without that guard, anyone later aligning the two would silently make the assertion vacuous.

Both directions have a negative control:

- reverting the builder to `TenantRef` fails with `expected ["tenant-ns"], actual ["tenant"]`
- renaming the fixture namespace to `"tenant"` fails the guard with `Should not be: "tenant"`

`make check` passes, `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` in `e2e/` reports 0 issues, and the 13 tests still run in 0.01s with no cluster. All three commits build on their own.

e2e/internal/fixture/fixture_test.go
```diff
@@ -0,0 +1,189 @@
+// Copyright © 2025 Kube logging authors
```
csatib02: The header is wrong on all files.

e2e/internal/fixture/fixture_test.go
```diff
@@ -0,0 +1,189 @@
...
+	f := Flow("ns", "test-flow", labels, out.Name)
+	require.Equal(t, labels, f.Spec.Match[0].Select.Labels)
+	require.Equal(t, []string{"test-output"}, f.Spec.LocalOutputRefs)
+}
+
+// The §6 safety net: build the Logging that fluentd-aggregator's MultiWorker
+// test writes by hand and assert the builder produces exactly it. A default
+// that quietly absorbed a divergence would show up here rather than after a
+// twenty minute cluster run.
+func TestBuilderReproducesTheMultiWorkerLiteral(t *testing.T) {
```
csatib02: We have no context about the document, as someone just looking at the code, keep comments like these out of the code.

e2e/internal/fixture/images.go
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fixture
+
+import "github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
+
+// The locally built images the suites load into each cluster. These are
+// duplicated from common while both exist; images_test.go asserts they stay
+// equal, so the two cannot drift the way the syslog-ng reloader name did.
```
csatib02: Please do a full sweep, and tell your agents to keep the comments to the absolute necessary amount, the code should speak for itself, we need comments if the code needs extra context.

csatib02 (commented):
Please sweep the code and clean it up as much as you can, there is a lot of unnecessary commenting going on.

e2e/internal/fixture/fixture_test.go
vyncint: Fixed in 38b8874. All six said `© 2025` but were written this year; the files added to the repo in 2026 all use `© 2026 Kube logging authors`, so that is what they say now.

`licensei header` ignores the year, which is why `make check` stayed green on the wrong one — so this was only ever going to be caught by review. Worth flagging that `e2e/internal/wait/conditions_test.go` carries the same stale `© 2025` from #2298; it is outside this PR's files, so I have left it alone, but I am happy to send the one-line fix separately.

e2e/internal/fixture/fixture_test.go
vyncint: Fixed in 38b8874. Fair point — the section numbers only meant something to whoever had the plan open, and nobody reading this file later will. Every reference to it is gone: no `§`, no issue numbers, and none of the "counted at N call sites" figures that were really just notes to myself.

The comment that was here is now one line stating the single intentional difference between the expected object and the literal the suite writes by hand, which is the part the code genuinely cannot say on its own.

e2e/internal/fixture/images.go
vyncint: Swept all six files in 38b8874: 78 non-header comment lines down to 46, and what is left is mostly doc comments on exported identifiers.

What went: the narrative comments explaining why the package exists, the per-suite counts behind each default, and the assertion messages that only restated the expected value — `require` prints that on failure anyway. What stayed is the handful that explains something the code cannot: why `WatchNamespaces` takes a namespace and not the ref, why `tenancyFluentdSpec` is not `FluentdSpec()`, and why two assertions use literals instead of the constants they are checking.

This comment was also wrong on its own terms: it pointed at `images_test.go`, which does not exist. The test is `TestImageNamesMatchCommon` in `fixture_test.go`, and it now says so.

e2e/internal/fixture/logging.go
```diff
@@ -0,0 +1,128 @@
...
+		Spec: v1beta1.LoggingSpec{
+			EnableRecreateWorkloadOnImmutableFieldChange: true,
+			ControlNamespace: namespace,
+		},
+	}
+	for _, o := range opts {
+		o(l)
+	}
+	return l
+}
```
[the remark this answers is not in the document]
vyncint: Fixed in e5171ce — the conclusion is right, the stated failure mode is not.

`Logging` is `scope: Cluster`, the tenancy builders here already omitted `ObjectMeta.Namespace`, and the base builder was the odd one out, so it no longer sets it. The first parameter is now named `controlNamespace` to say what it actually sets.

It does not cause a rejection, though. In `k8s.io/apiserver/pkg/registry/rest/meta.go`, `EnsureObjectNamespaceMatchesRequestNamespace` reaches `case requestNamespace == metav1.NamespaceNone` for a cluster-scoped kind and clears the field, returning nil — so there is no `metadata.namespace: Forbidden` to hit. That matches the evidence: `fluentd-aggregator` and the other suites have set it on their `Logging` literals for years and create them against real KIND clusters without complaint.

So this was inert noise rather than a break. Still worth removing, because a builder that sets a field the API server throws away invites the next reader to copy it.

csatib02:
Seems like the ci is failing :/

csatib02 (approved):

PR #2301, opened 2026-07-30
fix(e2e): give every kind cluster its own kubeconfig

### LINKED ISSUE #2287: E2E: unbounded kind CLI calls turn a stalled cluster build into a 20-minute panic that takes the whole package down

vyncint:
`kind delete` exits 1 under concurrency. kind takes a non-blocking lock on the kubeconfig it updates, so clusters sharing the default one lose the race and fail *after* the test body has passed. 96 concurrent deletes of clusters that do not exist: 7 fail with `failed to lock config file` on a shared kubeconfig, 0 when each names its own. Giving every cluster its own `--kubeconfig` also stops the suite rewriting the developer's `~/.kube/config`. The delete path discarded kind's stderr, which is why this only ever surfaced as `exit status 1`; it now reports what kind said.

### Rebased on #2302

The Makefile change came out in the rebase, since #2302 caps concurrency with `-parallel ${E2E_SUITE_PARALLEL}`, and is back on your note. The two are not the same bound: `-parallel` limits tests inside a binary, `-p ${E2E_CLUSTERS}` limits the binaries, and `-p` still defaults to `GOMAXPROCS`. Measured on a harness with this suite's package and test shape, peak clusters is 5 at four cores, 10 at eight and 15 at sixteen; pinning `-p` holds it at 5 everywhere. Both knobs are in `CLAUDE.md` now, beside the ones #2302 documented.

One correction to what I wrote above earlier: the peak is not `E2E_CLUSTERS` times `E2E_SUITE_PARALLEL`. That product is 8, but ten of the thirteen suites hold a single test, so the measured ceiling is 5.

Two conflicts resolved by hand, both in `KindClusterKubeconfig`:

- #2302's recreate-the-leftover delete now carries the per-cluster kubeconfig. Without it that call falls back to the shared default, and it fires exactly when clusters are colliding, which is the case this PR exists for.
- The create options moved into the `create` variable #2302 introduced, so the retry reuses them.

### Also from review

- **The partial-cluster delete lost the kubeconfig.** The delete that clears a half-built cluster after a create times out did not pass `options.Kubeconfig`, so it fell back to the shared default — and it runs exactly when several creates are timing out together.
- **Nothing removed the per-cluster kubeconfig, and its path was guessable.** `kindCluster.Cleanup` only removes the `os.CreateTemp` copy, and kind rewrites the file it manages rather than unlinking it, so a killed run handed its leftovers and any stale `.lock` to the next one. The files now live in one `os.MkdirTemp` directory per run at 0700, and `DeleteTestCluster` removes the kubeconfig and lock once the cluster is gone.
- **`RequireNoError` off the test goroutine.** `WithCluster` called it inside the `cluster.Start` goroutine, where `t.FailNow` is undefined, so a cluster that never started was dropped and resurfaced as an unrelated `Eventually` timeout. Start's result now comes back over a channel and is asserted in the deferred block.
- **`RequireNoError` on the teardown delete.** A cluster that came down cleanly still failed its test when kind's kubeconfig bookkeeping errored. It matches the `assert.NoError` calls above it now.
- **The test sharding was dead and inverted.** `sequence` is a per-binary counter and ten of the thirteen suites hold exactly one test, so with `SHARDS=2` shard 0 would run nothing and shard 1 everything. Gone, along with the commented-out matrix and its env vars; `Initialize` is the bare `t.Parallel`.
- **`helpers.go` hardcoded the tenant watch namespace.** `LoggingTenant` honours `nsTenant` everywhere else. Latent, since both callers pass `"tenant"`, but it left the live path disagreeing with the `internal/fixture` copy.
- **The kind stub tests are gone**, as requested — the shell stub raced and the flake landed on unrelated PRs. The timeout arithmetic keeps its five pure-Go tests. **`TestWatchNamespaces`** recomputes its expectation on every poll instead of capturing it once.

### Verification

`make check` passes, `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` in `e2e/` reports 0 issues, and all thirteen commits build, vet and test on their own after the rebase. Negative controls, re-run on the rebased head: putting the kubeconfigs back in the shared temp directory fails `TestClusterKubeconfigPath` on `Should not be: "/tmp"`, and skipping the lock removal fails `TestRemoveClusterKubeconfig` on `file ".../kind-gamma.kubeconfig.lock" exists`.

One thing your two requests collided on, worth flagging rather than quietly leaving. The pin you asked for on the partial-cluster delete lived in `TestInvocations/a stalled create deletes the partial cluster`, and that went with the stub tests you later asked to drop. The fix itself is still in `commands.go`, but nothing covers it now, and the same is true of the stderr capture. Both are cheap to re-pin without a shell stub if you want them held; say the word and I will, otherwise they ride on the real binary like the rest of the CLI plumbing.

On the evidence, unchanged: the stderr capture that would name a delete failure ships in the same commit as the fix, so the lock race is still inferred from the local experiment rather than observed in CI.


Makefile
```diff
@@ -242,7 +247,7 @@ test-e2e-nodeps:
 		KIND_IMAGE="$(KIND_IMAGE)" \
 		PROJECT_DIR="$(PWD)" \
 		E2E_TEST_COV_DIR=${TEST_COV_DIR} \
-		go test -count=1 -v -timeout ${E2E_TEST_TIMEOUT} $$(go list ./${E2E_TEST}/... | grep -vE '/e2e/(common|internal)(/|$$)')
+		go test -count=1 -v -p ${E2E_CLUSTERS} -parallel 1 -timeout ${E2E_TEST_TIMEOUT} $$(go list ./${E2E_TEST}/... | grep -vE '/e2e/(common|internal)(/|$$)')
```
csatib02: `-parallel 1` holds each binary to one test at a time, but `E2E_TEST_TIMEOUT` is per binary and still 20m. `fluentd-aggregator` has four `t.Parallel()` tests. Serialised they took 629.27s and 490.66s in 30590989952; parallel they took 160.87s and 269.18s in 30546098711. Headroom against the 1200s cap goes from roughly 7.5x to roughly 1.9x, and run-to-run variance on packages this change cannot affect is 2.7x.

The second effect is on #2288's own mechanism. `timeoutFraction = 0.8` derives the 16m kind cap from this same 20m. Once two tests have run one after the other, under 16m of package budget is left, so the cap can no longer fire before the binary panics. For the 3rd and 4th test the readable `timed out after 16m0s` error is unreachable and you get `panic: test timed out after 20m0s` instead.

Two ways out, either is fine:

- raise `E2E_TEST_TIMEOUT` to ~45m in this commit, and derive the kind cap from `t.Deadline()` instead of the package flag so it tracks what is actually left, or
- drop `-parallel 1` and keep `-p ${E2E_CLUSTERS}`. `-p` alone already stops peak clusters scaling with core count, which is the problem in the description, and it costs no wall time on any package.

e2e/internal/kind/commands.go
```diff
@@ -133,7 +133,14 @@ func (k *Kind) DeleteCluster(options DeleteClusterOptions) error {
 }
 
 func (k *Kind) deleteCluster(timeout time.Duration, options DeleteClusterOptions) error {
-	return k.run(timeout, options.AppendToArgs([]string{"delete", "cluster"}), nil)
+	cmderr := &bytes.Buffer{}
```
csatib02, line 139: Capturing kind's stderr here is the piece that was missing. `exit status 1` on its own is unattributable, which is why the delete failures in 30554200418 still cannot be pinned to the lock race.

One gap: the cleanup delete at line 113 does not pass the kubeconfig through.

```go
cleanupErr := k.deleteCluster(k.CleanupTimeout, DeleteClusterOptions{Name: options.Name})
```

That call fires when a create times out, which is exactly when several clusters are failing together, so it is the delete most likely to lose the lock. It should carry `options.Kubeconfig`. The `a stalled create deletes the partial cluster` case at commands_test.go:270 creates without a kubeconfig, so it does not catch this; setting one there pins it.

e2e/common/kind.go
```diff
@@ -27,10 +29,19 @@ const KindClusterCreationTimeout = "3m"
 
 var kindCLI = kind.New()
 
+// ClusterKubeconfigPath keeps each cluster's kind bookkeeping in its own file.
+// kind locks the kubeconfig it updates and the lock is non-blocking, so
+// clusters sharing one fail outright rather than wait. Cluster names are
+// already restricted to what kind accepts, which is safe in a path.
+func ClusterKubeconfigPath(name string) string {
+	return filepath.Join(os.TempDir(), "kind-"+name+".kubeconfig")
```
csatib02: Two things about this path.

Nothing removes the file. `kindCluster.Cleanup()` (cluster.go:167) removes only the `os.CreateTemp` copy made at cluster.go:91, and kind's `RemoveKIND` rewrites this file rather than unlinking it. The 17 cluster names are fixed so it will not grow, but a file left behind by a killed run is reused by the next one, along with a stale `.lock` next to it if kind was killed mid-write.

The path is also predictable in a shared directory, while `GetTestCluster` ten lines further down already uses `os.CreateTemp` for the same kind of file. `os.WriteFile` follows symlinks and will not re-chmod an existing file, so a pre-planted `/tmp/kind-fluentd-aggregator.kubeconfig` survives and gets written through.

One `os.MkdirTemp("", "e2e-kubeconfig-*")` per run at 0700, files inside it, directory removed at teardown, covers both and matches what this file does already.

csatib02 (commented):
Chased the flake through the CI history before this lands. The per-cluster `--kubeconfig` and the stderr capture look right. Two things on the rest.

**`-parallel 1` and the 16m kind cap work against each other.** `E2E_TEST_TIMEOUT` is unchanged at 20m and it applies per binary, so serialising `fluentd-aggregator`'s four tests spends that one budget four times over. From your own runs: 160.87s and 269.18s parallel in 30546098711, against 629.27s and 490.66s serialised in 30590989952. That is 41-52% of the 1200s cap where it was 13-22%. On the same green runner the single-test packages came in 5-10% faster than baseline (watch-selector 96.6s vs 106.9s), so the jump is serialisation rather than a slow runner. Detail inline on the Makefile.

**Most of this PR works around one line.** `e2e/common/cluster.go:73` uses `RequireNoError` for the teardown delete, so a cluster that was destroyed cleanly still fails the test when kind's kubeconfig bookkeeping errors. Lines 70 and 71 already use `assert.NoError`. In 30554200418, `TestFluentbitAgentDedicatedNamespace` logged `Collecting coverage files ...` and then failed only at `cluster.go:73`: the body passed and teardown killed it. `TestFluentbitSingleTenantPlusInfra` has the same shape, and the two are about 20s apart, so it is not a herd effect. Making 73 match 70-71 retires that failure class whatever the kubeconfig layout is.

Worth splitting in two. Land the per-cluster `--kubeconfig` and the stderr capture on their own, hold `-parallel 1`.

One note on the evidence. The stderr capture that would name the delete failure ships in the same commit as the fix, so 30589321074 still shows a bare `exit status 1`; the lock race is inferred from your local experiment rather than observed in CI. And the green run is one sample. Master was green on 4 of 5 runs after the series with the old flags, so it does not yet separate a real improvement from a quiet runner.

csatib02:
Three things in `e2e/common/helpers.go`. Putting them here rather than in the fixture follow-up because they all sit on the concurrency path this PR changes.

**`Initialize` is where `-parallel 1` actually lands.** helpers.go:60-70 does a sharding skip and then calls `t.Parallel()`, so that is the call the new flag throttles.

The sharding wrapped around it is dead and inverted. `localSeq` is a per-binary counter and 10 of the 13 suites hold exactly one test, so `localSeq` is always 1 in those binaries. With `SHARDS=2` that gives `1 % 2 != 0` for shard 0, which skips, and `1 % 2 == 1` for shard 1, which runs: shard 0 runs nothing and shard 1 runs everything. The matrix that would drive it is commented out (e2e.yaml:80-81, 136-137), so nothing exercises it either way. Either drop it and keep the bare `t.Parallel()`, or shard on `t.Name()` instead of a counter that cannot tell single-test packages apart. Dropping looks right given `E2E_CLUSTERS` now covers what sharding was for, and it makes the `-parallel 1` decision easier to reason about.

**`RequireNoError` calls `t.FailNow()` (helpers.go:53-58) and cluster.go:65-66 calls it off the test goroutine.**

```go
go func() {
	RequireNoError(t, cluster.Start(ctx))
}()
```

`FailNow` has to run on the test goroutine; anywhere else it is undefined and the error is dropped, so a cluster that fails to start surfaces later as a confusing `Eventually` timeout. Same class as `cluster.go:73` in my inline comment, in a file this PR already touches. #2295 fixed this shape for `LogProducer` and this one was left behind.

**helpers.go:231 still hardcodes what 96b31314 fixed in the other copy.** `LoggingTenant` takes `nsTenant` and honours it at :189, :205 and :230, then hardcodes the watch namespace:

```go
WatchNamespaces: []string{"tenant"},
```

Latent rather than live: both callers pass `nsTenant := "tenant"` (fluentbit-multitenant:77, fluentbit-hotreload:79), so it happens to line up today. But `internal/fixture/tenancy.go:116` already carries the corrected `WatchNamespaces: []string{nsTenant}`, so the live path and the fixture copy disagree on the same logic. If the fixture work lands separately, worth carrying that one line back here so the fix is not blocked on it.

csatib02:
Please stop your agent from over-commenting. :)

e2e/common/helpers.go
```diff
@@ -58,14 +53,6 @@ func RequireNoError(t *testing.T, err error) {
 }
 
 func Initialize(t *testing.T) {
-	localSeq := atomic.AddUint32(&sequence, 1)
```
csatib02, line 61: Sharding has some residue env-vars set in the ci too, please clean that up too.

Makefile
vyncint: Took the second option — `-parallel 1` is gone in a314e11, `-p ${E2E_CLUSTERS}` stays.

The second effect you describe is the decisive one. The 41-52% headroom against 2.7x run-to-run variance is arguable on its own, but silently disarming the 16m cap that this same series added is not: for the 3rd and 4th test the readable `timed out after 16m0s` becomes unreachable and the binary panics with `test timed out after 20m0s`, which is the failure mode the cap existed to remove. Trading a legible error for a slightly lower cluster peak is the wrong way round.

Raising `E2E_TEST_TIMEOUT` and deriving the cap from `t.Deadline()` is the better shape and I would rather send it separately, with timings for the serialised layout rather than inferred from the parallel ones.

One thing I should be straight about now that the flag is gone: `-p 4` is close to a no-op on today's four core runners, since `go test` already defaults `-p` to `GOMAXPROCS`. It stops the peak growing if runners get bigger, and it puts the bound in one named place, but it is a guard rather than a fix on the runners we have. The ceiling it gives is 8 by arithmetic — `fluentd-aggregator`'s 4, `volumedrain`'s 2 and one each from two more — against 17 unbounded. Happy to drop the Makefile change entirely if you would rather this PR were only the kubeconfig and stderr work.

e2e/internal/kind/commands.go
vyncint: Fixed in db550bb0. The cleanup delete carries `options.Kubeconfig` now, and you are right that it is the one that matters most: it fires when creates are timing out together, which is precisely when a shared kubeconfig loses the lock.

Setting a kubeconfig on the `a stalled create deletes the partial cluster` case does pin it, as you said. Without the fix it now fails on the arguments rather than passing:

```
expected: delete cluster --kubeconfig /tmp/kind-stuck.kubeconfig --name stuck
actual  : delete cluster --name stuck
```

Before the change the case only asserted that a delete happened at all, so it went green either way.

e2e/common/kind.go
vyncint: Fixed in 249d83b2, taking the `os.MkdirTemp` suggestion as written.

One directory per run from `os.MkdirTemp("", "e2e-kubeconfig-*")`, which is 0700 by construction, with the per-cluster names inside it. The file names stay predictable and the directory is not, so there is no longer a guessable path in the shared temp directory to plant anything at. `DeleteTestCluster` removes the kubeconfig and the `.lock` beside it once the cluster is gone, and the directory goes with the last cluster of the binary — `os.Remove` is a no-op on it while others remain.

`common` had no test files at all, so there was nothing pinning any of this. It has three now, and both halves have a negative control:

- putting the path back in `os.TempDir()` fails on `Should not be: "/tmp"`
- dropping the lock removal fails on `file "/tmp/kind-gamma.kubeconfig.lock" exists`

They also pin the path being stable across calls, which is what create and delete depend on to name the same file.

vyncint:
All three fixed. Checked each against the code first rather than taking them on trust, and all three hold exactly as described.

**Sharding — aa54d519.** Dead and inverted, confirmed. `sequence` is per binary, and I counted the suites: ten of thirteen hold exactly one test, so `localSeq` is always 1 in those. With `SHARDS=2` shard 0 skips on `1 % 2 != 0` and shard 1 runs on `1 % 2 == 1`, so shard 0 would have run nothing. The matrix is commented out at both places you cite. Dropped rather than reshuffled onto `t.Name()`, since `E2E_CLUSTERS` covers what it was for, and `Initialize` is now the bare `t.Parallel`.

**`RequireNoError` off the test goroutine — 8e5a84fa.** `cluster.Start` blocks, so unlike #2295's `LogProducer` this one cannot simply lose its `go`. Start's result comes back over a buffered channel and is asserted in the deferred block instead, on the test goroutine and after `cancel()`. That also stops the goroutine outliving the test, which was a second way the same call could misbehave. I bounded the wait at a minute so a runnable that will not stop gets named rather than running the binary out of its `-timeout` — swapping one unreadable panic for another would not be much of a fix.

The same commit makes the teardown delete `assert.NoError`, matching lines 70-71. Your reading of `30554200418` is what convinced me it is worth doing independently of the kubeconfig layout: the body passed, coverage was collected, and only `cluster.go:73` failed. `FailNow` there also cut the rest of the deferred block short, so it was losing cleanup as well as reporting a false failure.

**`helpers.go:231` — 10ce2fac.** Carried back, one line, so the live path and `internal/fixture` no longer disagree. Latent as you say — both callers pass `nsTenant := "tenant"` — so nothing changes for the suites as they stand.

`make check` passes, `golangci-lint` in `e2e/` reports 0 issues, and all eight commits build, vet and test on their own.

e2e/common/helpers.go
vyncint: Gone in bc808c20, both blocks. `fail-fast` stays — it came in with an earlier k8s-version matrix, not this one, so say the word if you want it out too.

vyncint:
Fair, and noted for the replies as much as the code. Trimmed in a44ce1e8: 27 lines of comment prose down to 8, and the assertion messages that only restated the expected value are gone.

csatib02:
Re-ran the numbers on the final head. This one looks done: `fluentd-aggregator` is back to 272.5s against 629.3s and 490.7s under `-parallel 1`, so the headroom is restored, and the teardown and kubeconfig changes cover what I raised. The self-review in 55a39155 also caught a removal race I had read past.

One thing left before the baseline is clean, and it is not this PR's doing. `elasticsearch-multiversion` is now the only package near its cap. Six runs:

| run | result | duration |
| --- | --- | --- |
| 30590989952 | ok | 387.9s |
| 30546098711 | ok | 625.5s |
| 30625716801 | ok | 856.4s |
| 30622235964 | ok | 988.5s |
| 30621489940 | FAIL | 1200.0s |
| 30554200418 | FAIL | 1200.0s |

A 3.1x spread across the green runs, 988.5s is 82% of the 1200s cap, and two of the six hit it. `30621489940` is on this branch, so it is not a master-only effect. Both failures are `panic: test timed out after 20m0s`, which is the one failure mode that skips every deferred teardown: no stern dump, no coverage, and the kind cluster stays up for the rest of the job.

Two things feed it.

`esReadyMargin` is 30s and the comment at :53-56 already says it does not cover teardown. Teardown is a stern dump, a `kubectl exec` plus `tar` for coverage, and a `kind delete`, which together run well past 30s.

The three waits at :490, :495 and :500 each call `esReadyBudget(t)` separately, and each call re-derives `time.Until(deadline) - 30s`. So the first wait is allowed to consume everything, and once under the margin `budgetWithin` floors at `time.Second`, so the remaining two fail almost at once. The reported failure is then `elasticsearch8` or `elasticsearch9` not ready when the actual cause is that `elasticsearch7` spent the package budget.

Cheapest first, if you want to pick this up separately:

1. Divide the budget rather than re-deriving it whole. Something like `remaining / wavesLeft` so one slow deployment cannot starve the other two, and the failure names the deployment that was actually slow.
2. Give this package its own `-timeout`. Each suite is its own binary, so it can run at 30m while the rest stay at 20m, without touching the others.
3. The root cause is that readiness is most of this package's runtime. A liveness-probe kill-loop has been the suspicion for a while but nobody has captured a restart count. One `kubectl get pod -o jsonpath='{..restartCount}'` inside the wait loop would settle it, and if it is the probe then the budget arithmetic stops mattering.

Not asking for any of this here. This PR is green and worth landing as it is; raising it so the next one starts from a baseline where a timeout means something is actually wrong.

vyncint:
Hi @csatib02, heads up that c3ee82bd3 landed just after your comment, so it is not in what you measured: no behaviour change, it just fixes the reasoning in 55a39155 (kind recreates the missing parent itself at 0755, so the mode is the real issue) and the test now checks the 0700. Happy to take the `elasticsearch-multiversion` budget problem in a separate issue once this ands. CI is green on the new head, so if that correction looks right, could you turn this into an approval?

csatib02:
The reasoning is right, the code does not follow it yet.

`os.MkdirAll` returns nil without touching the mode when the directory already exists, so the 0700 is not reasserted in the case this commit names. A throwaway test in `e2e/common` reproduces it:

```go
func TestProbe(t *testing.T) {
	first, err := clusterKubeconfigPath("probe-a")
	require.NoError(t, err)
	dir := filepath.Dir(first)

	require.NoError(t, os.RemoveAll(dir))
	require.NoError(t, os.MkdirAll(dir, 0o755)) // kind's doing, per this commit

	_, err = clusterKubeconfigPath("probe-b")
	require.NoError(t, err)

	info, err := os.Stat(dir)
	require.NoError(t, err)
	require.Equal(t, os.FileMode(0o700), info.Mode().Perm())
}
```

Gives `0755`.

`TestClusterKubeconfigPathRestoresTheDirectory` passes because `os.RemoveAll` leaves the directory absent, so `MkdirAll` creates it fresh at 0700. It never puts a 0755 directory in place, which is the state the fix exists for.

Two lines: `os.Chmod(dir, 0o700)` after the `MkdirAll`, and have the test pre-create at 0755 rather than removing it.

Low impact either way. `MkdirTemp` gives 0700, nothing in the suite removes the directory, so 0755 needs an outside actor, and the kubeconfigs themselves are 0600, so the directory exposes cluster names rather than credentials. I would still rather not land a test that passes for a different reason than the one it documents.

Separately: you checked v0.32.0, `Makefile:20` pins `KIND_VERSION ?= 0.31.0`. Worth confirming the 0755 behaviour on the version CI downloads.

Happy to approve with those two lines in. Everything else holds on c3ee82bd, 26 checks pass, and `fluentd-aggregator` at 272.5s confirms the headroom is back. ES budget as a separate issue works for me.

csatib02:
Great!

A few things I would rather handle now, before we move forward:

1. elasticsearch-multiversion budget. As discussed above, a separate issue works. It is the only package near its cap, and it decides whether a red e2e run means anything.
2. The tests with embedded shell scripts inside, I would instead drop those, we don't get much benefit from them anyway, and they are flaky.
3. TestWatchNamespaces in controllers/logging. ci.yaml run 30621399293 on this branch failed on it, --- FAIL: TestWatchNamespaces (10.01s). Let's tackle that now, if we sensed it!

vyncint:
@csatib02 both done in 2d855bf6 and c30dc38d. Ready for another look when you have a moment, and happy to get your approval if it all reads right.

csatib02:
Hey @vyncint,

Please rebase this PR, and resolve the conflicts!

vyncint:
Hi @csatib02, I rebased this PR, and resolved the conflicts. Please help me review it again!

csatib02:
One thing lost in your rebase.
The -p ${E2E_CLUSTERS} is gone, therefore the `E2E_CLUSTERS count = 0` in Makefile.
Line 250 now only does `-parallel ${E2E_SUITE_PARALLEL}`.
Please re-add it, then this is good to go!

csatib02 (approved):

PR #2307, opened 2026-08-04
fix(e2e): share the elasticsearch readiness budget between the waits

### LINKED ISSUE #2305: e2e: elasticsearch readiness budget is spent by the first deployment
### What

`elasticsearch-multiversion` waits for three deployments in a loop, and each iteration re-derives the whole remaining package budget:

```go
for _, name := range []string{"elasticsearch7", "elasticsearch8", "elasticsearch9"} {
    require.Eventually(t, func() bool { ... }, esReadyBudget(t), 10*time.Second)
}
```

`esReadyBudget` returns `time.Until(deadline) - 30s`, so the first iteration is allowed to spend everything. Once it has, `budgetWithin` floors the next two at one second and they fail almost immediately. The reported failure then names `elasticsearch8` or `elasticsearch9` when the deployment that actually stalled was `elasticsearch7`.

`esReadyMargin` is 30s and its own comment says it does not cover teardown — a stern dump, a `kubectl exec` plus `tar` for coverage, and a `kind delete` run well past that.

### What changed since this was last measured

Worth recording before anyone acts on the older numbers. The 3.1x spread and the two runs at the 1200s cap were measured before #2302 landed the liveness-probe fix. Six runs since:

| run | result | duration |
| --- | --- | --- |
| 30818421306 | ok | 356.4s |
| 30804341601 | ok | 379.0s |
| 30859321423 | ok | 368.5s |
| 30893846956 | ok | 366.1s |
| 30861138335 | ok | 392.1s |
| 30859327501 | FAIL | 639.4s |

Five healthy runs sit in 356-392s, a 1.1x spread at about 30% of the cap. The one failure was on a degraded runner: total suite work was 4074s against roughly 2600s for the others, and two suites this branch does not touch failed in the same run.

So this is no longer a suite that runs near its ceiling. What is left is an attribution bug — a slow first deployment blames the second — and a margin that does not cover teardown. Both are cheap and worth fixing; neither is urgent in the way the earlier numbers suggested.

### Suggested

- Divide the remaining budget across the deployments still to check rather than handing all of it to the first, so a stall names the deployment that caused it.
- Raise the margin to something that covers teardown, or derive it the way `internal/kind` derives its command timeout.

### Why it matters for the migration

`internal/harness` now owns wait budgets for migrated suites and caps a single wait at `min(5m, remaining - 90s)`. Whatever this issue settles should decide whether the harness grows a per-call budget override, since this suite is the one case where the shared default may not fit.

vyncint:
Closes the attribution half of #2305.

`esReadyBudget` returned `time.Until(deadline) - 30s` and the loop called it fresh for each of the three deployments, so the first was allowed to spend everything. Once it had, `budgetWithin` floored the next two at a second and they failed almost at once — under the wrong deployment's name. It divides by the waits still to come now, and the assertion says which deployment never came up.

The margin goes from 30s to 90s. Its own comment said 30s did not cover teardown, which is a stern dump, a `kubectl exec` plus `tar` for coverage, and a `kind delete`. 90s is also what `internal/harness` reserves, which is one less number to reconcile when the three budget derivations in this module get collapsed.

The fallback is not divided: with no deadline there is no shared budget to split, and it is already per-wait.

### Verification

`make check` passes and `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` in `e2e/` reports 0 issues.

`TestBudgetWithin` gains the count of remaining waits, and `TestBudgetLeavesRoomForTheRemainingWaits` pins what the split is for: three times the first budget still fits inside what was left. Negative control: restoring `remaining - esReadyMargin` fails it.

The floor is applied after the division, not before. Dividing first made the effective minimum `1s / waitsLeft` — `budgetWithin(esReadyMargin+1s, 3)` returned 333ms, positive enough to skip the floor, and a budget that small expires on the first tick and names `elasticsearch7` again. That is a second per wait rather than a second shared between them, so above the floor the invariant above holds and at the floor it does not; a package that close to its deadline is finished either way.

On a real cluster the three-deployment ladder works — `elasticsearch7`, `8` and `9` each became ready, so nothing starved. The suite then fails later on the ES7 document count, and so does master's on the same machine: 497s against 507s, both polling `0` eighteen times, both at the same assertion. That is a log-delivery problem local to this machine rather than anything in this change, and CI has the suite green.

### DIFF
--- a/e2e/elasticsearch-multiversion/elasticsearch_multiversion_test.go
+++ b/e2e/elasticsearch-multiversion/elasticsearch_multiversion_test.go
@@ -50,9 +50,10 @@ const (
 	// esReadyFallback applies when the binary has no deadline to derive from.
 	esReadyFallback = 15 * time.Minute
 
-	// esReadyMargin is the room left before the deadline for the wait to report
-	// its own failure. It does not cover teardown, which can outlast it.
-	esReadyMargin = 30 * time.Second
+	// esReadyMargin is the room left before the deadline. 30s covered the wait
+	// reporting its own failure but not teardown, which is a stern dump, a
+	// kubectl exec plus tar for coverage, and a kind delete.
+	esReadyMargin = 90 * time.Second
 )
 
 var TestTempDir string
@@ -93,21 +94,22 @@ func logContainerRestarts(t *testing.T, c common.Cluster, ctx context.Context, n
 	}
 }
 
-// esReadyBudget caps a wait strictly below the enclosing -timeout, so an
-// Elasticsearch deployment that never becomes ready fails by name instead of
-// letting the package binary panic and report only a goroutine dump.
-func esReadyBudget(t *testing.T) time.Duration {
+// esReadyBudget shares what is left of the -timeout between the deployments
+// still to check. Handing it all to the first let a slow one spend the
+// package's budget, leaving the rest to fail instantly under the wrong name.
+// The fallback is already per-wait, so it is not divided.
+func esReadyBudget(t *testing.T, waitsLeft int) time.Duration {
 	deadline, ok := t.Deadline()
 	if !ok {
 		return esReadyFallback
 	}
-	return budgetWithin(time.Until(deadline))
+	return budgetWithin(time.Until(deadline), waitsLeft)
 }
 
 // budgetWithin never returns a non-positive duration: require.Eventually treats
 // one as already expired, which would fail every run instead of only stalled ones.
-func budgetWithin(remaining time.Duration) time.Duration {
-	if budget := remaining - esReadyMargin; budget > 0 {
+func budgetWithin(remaining time.Duration, waitsLeft int) time.Duration {
+	if budget := (remaining - esReadyMargin) / time.Duration(waitsLeft); budget > 0 {
 		return budget
 	}
 	return time.Second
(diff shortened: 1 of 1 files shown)


csatib02:
Correct fix, and the right half of #2305 to do first. One boundary note.

Dividing re-derived `time.Until(deadline)` each iteration rather than fixing a static third is better than what the issue asked for: a fast `elasticsearch7` hands its slack to `elasticsearch8` instead of it being lost. `len(deployments)-i` also keeps `waitsLeft` at 1 or more, so the division has no zero case to guard from the only call site. Moving the name into `Eventuallyf` is the part that actually settles the attribution — a scrolled-away `t.Logf` was the problem.

`TestBudgetLeavesRoomForTheRemainingWaits` asserting `3*first <= remaining - esReadyMargin` is the invariant rather than the arithmetic, which is what makes it worth keeping.

**The floor now sits below a second.** This case went away rather than getting a `waitsLeft`:

```go
{"just above the margin", esReadyMargin + time.Second, time.Second},
```

It could not be updated, and the reason is the finding: `budgetWithin(esReadyMargin+time.Second, 3)` is `1s/3`, so 333ms. That is positive, so it skips the floor and is returned. The comment above `budgetWithin` still says the floor exists because "require.Eventually treats one as already expired", but dividing after the subtraction and before the floor makes the effective minimum `1s / waitsLeft` rather than 1s.

A 333ms budget fails on the first tick and reports `elasticsearch7 never became ready`, which is the misattribution this PR removes, reappearing at the boundary.

Low severity — it needs `time.Until(deadline)` to be within a few seconds of the 90s margin, and a package that close to its deadline is finished either way. But it is a one-line fix, floor after dividing rather than before:

```go
if budget := (remaining - esReadyMargin) / time.Duration(waitsLeft); budget > time.Second {
    return budget
}
return time.Second
```

csatib02 (approved):

PR #2321, opened 2026-08-19
fix(e2e): retry a wedged cluster delete instead of failing the suite

vyncint:
Follow-up to the teardown failure seen on #2308.

Docker intermittently reports that it killed a kind node container but never saw it exit:

```
cannot remove container "/X-control-plane": could not kill: tried to kill container, but did not receive an exit event
```

`kind delete cluster` then exits 1, and the teardown asserted on that. In run 32119183436 it hit five clusters at three separate points and turned **8 of 13 suites red** — while every assertion in them passed and no log delivery failed anywhere in the run. Eight of the nine failures originate at the `DeleteTestCluster` assert, and two of the failing suites were not touched by that PR at all.

### Retry

`DeleteCluster` makes three attempts with a 5s backoff. That is safe because kind exits 0 deleting a cluster that is already gone, so repeating after a partial success cannot fail.

The delete inside `CreateCluster` stays single-shot: it runs on the short cleanup budget of an already-failing create, where the caller still has to report within what is left.

### Then log rather than assert

Once the retries are out, teardown reports the failure instead of failing the test. By then the assertions have decided the verdict, and failing there cannot remove the leftover anyway — the next run recreates one. Same call the coverage collection made in #2296.

`DeleteTestClusterOrLog` replaces `assert.NoError` at its four call sites.

### Verification

`make check` passes, and `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`.

The tests drive the real command path through a stub kind binary that counts invocations across exits, so the retry, the stop-on-first-success, the backoff, the last-error propagation and the zero-value case are exercised rather than mocked. They were checked against mutations instead of assumed to bite: forcing attempts to 1 fails three of them, and keeping the first error rather than the last fails a fourth.

Against a live cluster the delete still lands on the first attempt, so the healthy path pays nothing.

### DIFF
--- a/e2e/common/cluster.go
+++ b/e2e/common/cluster.go
@@ -67,7 +67,7 @@ func WithCluster(name string, t *testing.T, fn func(*testing.T, Cluster), before
 	if err != nil {
 		// The cluster is created before the client can fail, and the deferred
 		// teardown below is not registered yet.
-		assert.NoError(t, DeleteTestCluster(name))
+		DeleteTestClusterOrLog(t, name)
 	}
 	RequireNoError(t, err)
 
@@ -90,7 +90,7 @@ func WithCluster(name string, t *testing.T, fn func(*testing.T, Cluster), before
 			assert.Fail(t, "cluster.Start did not return after cancellation")
 		}
 
-		assert.NoError(t, DeleteTestCluster(name))
+		DeleteTestClusterOrLog(t, name)
 	}()
 
 	fn(t, cluster)
@@ -130,6 +130,15 @@ func GetTestCluster(clusterName string, opts ...cluster.Option) (Cluster, error)
 		kubeconfigFilePath: kubeconfigFile.Name(),
 		clusterName:        clusterName,
 	}, errors.WrapIfWithDetails(err, "creating cluster with rest config", "cfg", restCfg)
+}
+
+// A delete that fails after the assertions have run is the runner's state, and
+// failing here would not remove the leftover the next run recreates anyway.
+func DeleteTestClusterOrLog(t *testing.T, clusterName string) {
+	t.Helper()
+	if err := DeleteTestCluster(clusterName); err != nil {
+		t.Logf("deleting cluster %q: %v", clusterName, err)
+	}
 }
 
 func DeleteTestCluster(clusterName string) error {
--- a/e2e/internal/harness/harness.go
+++ b/e2e/internal/harness/harness.go
@@ -152,7 +152,7 @@ func (b *Builder) Start() *Env {
 	if err != nil {
 		// The kind cluster is up before the client can fail, and teardown is
 		// not registered yet.
-		assert.NoError(t, common.DeleteTestCluster(b.cfg.cluster))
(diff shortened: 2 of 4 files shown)


e2e/common/cluster_test.go
csatib02, line 1: Please remove this file, we are not interested in testing kinds behavior, and whether a log happens or not.

e2e/internal/kind/delete_test.go
csatib02, line 1: Delete this file too, same reason.

csatib02:
The diagnosis and the fix are right, and the evidence in run 32119183436 makes the case on its own: eight of nine failures at the same `DeleteTestCluster` assert, every assertion in those suites passing, and two of the failing suites untouched by the PR that surfaced it.

Three decisions worth calling out as correct. Retrying is safe because `kind delete` exits 0 on a cluster that is already gone, so repeating after a partial success cannot fail. Not retrying `ErrTimeout` is the right distinction — a wedged docker fails in seconds, a hung kind has already spent the budget the caller has left to report within. And leaving `CreateCluster`'s internal delete single-shot keeps the cleanup path on the short budget it needs.

There is also a live instance of that flake in it. `TestDeleteClusterDoesNotRetryATimeout` sets `CommandTimeout: 200 * time.Millisecond` against a stub that has to spawn `/bin/sh`, read and rewrite a counter file, and only then `sleep 30`. If the shell does not reach the counter inside 200ms the assertion sees `calls() == 0` and fails. That is the race that failed `TestInvocations/a_stalled_create_deletes_the_partial_cluster` 2 times in 12 on an idle machine and 3 in 3 on a loaded one, at the same 200ms. Every other new test uses `CommandTimeout: time.Minute` and is not exposed to it. Deleting the file resolves it; if any of it comes back later, that value is the thing to change.

Two of the deleted tests do encode real decisions rather than plumbing — that a timeout is not retried, and that a failed teardown does not change a passing verdict. Both are fine to lose here because both are visible in CI: if the retry stops working, the wedged-delete failures return in exactly the shape you already captured. Worth keeping the comment on the `ErrTimeout` branch though, since after the test files go it is the only remaining record of why that case returns instead of retrying.

`TestNewConfiguresTheDeleteRetry` asserting that `New` sets its own defaults would have been worth dropping regardless — it cannot fail for a reason a reader would not already assume.

vyncint:
@csatib02 Both files are gone — that push landed about a minute before you submitted, so the diff you commented on is stale.

One thing to flag rather than have you find it: I replaced them with `internal/kind/retry_test.go` instead of leaving nothing. It has no stub, no `exec` and no log assertion — the retry loop takes a `func() error`, so it tests our own control flow rather than kind's. That is the line `ed254505` drew when it dropped the stub tests and kept the timeout arithmetic. Five subtests, all instant, and the two that matter fail if the retry or the `ErrTimeout` guard is removed.

If you would rather have nothing there at all, say so and I will drop it.

csatib02 (approved):

PR #2322, opened 2026-08-19
refactor(e2e): take the images and the receiver URL from one place each

vyncint:
The two extractions the rule flagged on #2308, now that both prerequisites have landed.

### Images

`fluentd-full` was declared three times — `common/helpers.go`, `setup.defaultImages`, and `internal/fixture/images.go` — with the other two agreeing with the loader's copy by hand. `internal/image` is now the only declaration, and 125 reference sites resolve to it.

One deviation from what was sketched. They cannot sit on the harness: `harness` imports `common/setup`, which imports `common`, so the loader cannot import the harness back. Putting them there would strand `defaultImages` and leave two declarations instead of three, not one. `internal/image` sits below both instead, which is the only place that serves the loader and every suite, migrated or not. It folds into the harness when `common/setup` does.

`TestAllMatchesTheLoaderList` pins the six entries against the literals they replaced, so a drifting repository or env var fails there rather than loading one image and running another.

### Receiver URL

Nine sites across eight suites spelled `http://%s-test-receiver:8080/%s` inline while `fixture.ReceiverURL` had no caller outside `tenancy.go`. `Receiver` moves to its own file and gains `URL`, `URLIn` for a client in another namespace, and `ReceiverName`, which the log selector and both URLs now derive from instead of respelling the suffix.

The package functions take a release rather than an `Env` so the seven unmigrated suites can use them too; they go when the last one migrates, leaving the methods. `common`'s two copies stay, since `common` cannot import the harness, and they go with `common`.

### Worth noting

After images move out, all eight suites that #2308 swept stop importing `internal/fixture` altogether — images were the only thing they used it for. What remains is `tenancy.go` with two callers.

The inline URL count is 8 suites and 9 occurrences rather than ten; `volumedrain` has two of them.

### Verification

`make check` passes and `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`. Both commits build, vet and test on their own.

`TestBuildersMatchCommon` is the useful cross-check: `fixture` now computes the endpoint through the harness while `common` still uses its own inline literal, and the two objects stay identical.

Against real clusters, covering both the method form and the package form: `fluentbit-multitenant` 96s, `syslog-ng-aggregator` 131s, `watch-selector` 82s, `fluentbit-agent-namespace` 85s, the last of those being an unmigrated suite.

### Receiver log adoption — split out

`Receiver.Logs` moves here with only its own two callers while the unmigrated suites still shell out. Putting the inventory on the record rather than leaving it to be rediscovered:

| suite | total | `logs` | `exec` | `scale` |
| --- | --- | --- | --- | --- |
| `volumedrain` | 8 | 2 | 2 | 4 |
| `elasticsearch-multiversion` | 3 | 0 | 3 | 0 |
| `fluentbit-agent-namespace` | 1 | 1 | 0 | 0 |
| `fluentd-aggregator` | 1 | 1 | 0 | 0 |
| `fluentd-aggregator-detached` | 1 | 1 | 0 | 0 |
| `fluentd-aggregator-namespacelabel` | 1 | 1 | 0 | 0 |
| `logging_metrics_monitoring` | 1 | 0 | 1 | 0 |
| `syslog-ng-aggregator-detached` | 1 | 1 | 0 | 0 |
| **total** | **17** | **7** | **6** | **4** |

**Seven of the seventeen are in scope** — every `kubectl logs` call, all of them receiver reads selecting on `harness.ReceiverName`. The other ten are not: six are `kubectl exec` (coverage collection, and the elasticsearch curl pod) and four are `kubectl scale` on the receiver deployment, which is what `volumedrain` drains against.

The conversion is not purely mechanical, which is why it is separate: five of the seven pass `--tail 30` while `volumedrain`'s two read unbounded, so a shared helper has to decide that rather than inherit it, and bounding a read changes what a `Contains` check can see.


e2e/syslog-ng-aggregator-detached/syslog_ng_aggregator_detached_test.go
```diff
@@ -312,7 +313,7 @@ func TestSyslogNGDetachedIsRunningAndForwardingLogs(t *testing.T) {
 			cmd := common.CmdEnv(exec.Command("kubectl",
 				"logs",
 				"-n", ns,
-				"-l", fmt.Sprintf("%s=%s-test-receiver", types.NameLabel, releaseNameOverride)), c)
+				"-l", fmt.Sprintf("%s=%s", types.NameLabel, harness.ReceiverName(releaseNameOverride))), c)
```
csatib02, line 316: Shouldn't we use the new receiver abstraction here, this occurred in other tests as well.

csatib02:
Backing the two comments above with counts, since both are easier to act on with a number.

**Comments.** The diff adds 60 comment lines against 308 lines of code, so roughly one in five. That is about double the ratio of the recent e2e work, and this is the change least in need of it: moving a constant to one place and replacing nine inline literals with a call is self-evident from the diff. The ones worth keeping are the two that record a decision a reader cannot recover — why `internal/image` sits below the harness rather than on it, and why the package-level functions take a release instead of an `Env`. Most of the rest restate the statement underneath them.

**`Receiver.Logs`.** Eight suites still shell out to `kubectl` directly, 18 occurrences:

```
volumedrain                        8
elasticsearch-multiversion         4
fluentbit-agent-namespace          1
fluentd-aggregator                 1
fluentd-aggregator-detached        1
fluentd-aggregator-namespacelabel  1
logging_metrics_monitoring         1
syslog-ng-aggregator-detached      1
```

This is the same shape as the problem the PR fixes. `fixture.ReceiverURL` existed with no caller outside `tenancy.go` while nine sites spelled the URL inline, and the fix was to adopt it everywhere rather than leave it sitting there. Adding `Logs` and leaving the call sites behind sets up the same rediscovery next time.

Not all 18 are receiver logs — `volumedrain`'s eight and some of the elasticsearch four are `kubectl exec` and other subcommands that `Logs` cannot replace. Worth stating the number that can convert and either doing them here or opening the follow-up with that count in it, so the remainder is a decision on the record rather than something the next reader has to work out.

The rest of the PR stands: the three image declarations collapse to one, the import cycle makes `internal/image` the right home rather than the harness, and `TestAllMatchesTheLoaderList` is the test that stops a drifting env var loading one image while the suite runs another.

csatib02:
Re-measured on `4f1c23d4`, and I think this push crossed my comment rather than answering it — flagging it so it does not sit as though it were handled.

| | before | now |
| --- | --- | --- |
| comment lines added | 60 | 53 |
| code lines added | 308 | 308 |
| ratio | 1 in 5.1 | 1 in 5.8 |
| `kubectl` call sites | — | 1 removed, 1 added, net zero |
| `.Logs(` calls added | — | 2 |

So seven comment lines went and `watch-selector` joined the diff, but the narration ratio is where it was and no shell-out was actually replaced.

Either resolution works, I would just like it decided rather than left open:

1. Do both here — cut the narration to the two comments that record a decision (`internal/image` sitting below the harness, and the package functions taking a release), and convert the receiver-log call sites `Logs` can actually replace.
2. Split the adoption into a follow-up — in which case put the count in this PR's body and say which of the 18 sites are in scope, so the remainder is a recorded decision instead of something the next reader rediscovers.

My preference is 2 if the conversion is more than a handful of sites, since this PR is already a clean 3-declarations-to-1 story and the adoption is a separate concern. What I would rather avoid is `Logs` landing with two callers and eight suites still shelling out, because that is the same shape as `fixture.ReceiverURL` sitting unused while nine sites spelled the literal inline — which is the thing this PR exists to fix.

Nothing else outstanding: the import-cycle reasoning for `internal/image` holds, the three declarations do collapse to one, and `TestAllMatchesTheLoaderList` is the check that stops a drifting env var loading one image while the suite runs another.

For reference, #2321 is a good model for the comment question — `retryDelete` taking the delete as a parameter made the tests cheaper and removed the stub race, and it carries two comments total, both recording a decision the code cannot state.

csatib02:
Separate from the two open points, and worth fixing here since this PR is the first place it is visible in one file.

The `Env` override does not reach the specs. `Ref()`, `Spec()` and `Basic()` all build from `Repository` and `Tag`, while the loader is the only reader of `Env`:

```go
// Read by the loader only: an override loads one ref, the specs name this.
Env        string
```

So `FLUENTD_IMAGE=myrepo/fluentd:v2` loads that image into the cluster while every CR still names `fluentd-full:local`, and the aggregator fails to pull. The override only works while it happens to agree with the default, which is the case the Makefile passes.

This is not something the PR introduced — `setup.defaultImages` read `lookupEnv` for loading while the suites used `common.FluentdImageRepo` and `common.FluentdImageTag`, which ignored it. The split has been there the whole time. What changed is that both halves now live in `internal/image`, so one type owns the contradiction and it can be fixed in one place instead of two.

The fix I would expect is for the constructors to resolve the env var when it is set, so `Repository` and `Tag` carry the override and the loader keeps reading one field rather than the raw name. Then `Ref()` and `Spec()` agree by construction and `TestAllMatchesTheLoaderList` keeps pinning them against the loader list. Worth a case with the variable set, since that is the path nothing exercises today.

If you would rather not widen this PR, an issue is fine — but this is the one moment the two halves are in the same file, and it will cost more to fix once `common/setup` folds into the harness.

The copilot thread on `image.go:30` is describing this same behaviour from the doc-comment side. It is accurate about the divergence, and wrong that the comment misdocuments it — the comment says exactly what the code does. Closing it in favour of this.

csatib02:
The `Env` fix is right, and the two cases I would have asked for are already in it. `splitRef` taking `LastIndex` and then rejecting a `/` after the colon is the correct way to do this — `registry:5000/fluentd:v2` splits at the tag rather than the port — and `a registry port is not the tag separator` pins it, alongside the no-tag case. `Ref()`, `Spec()` and `Basic()` now agree with the loader by construction rather than by hand. That closes the point and the copilot thread on `image.go:30` with it.

The other two are where I am stuck, so I would rather ask directly than keep measuring.

Re-measured on `b9915caa`:

| | first ask | now |
| --- | --- | --- |
| comment lines added | 60 | 60 |
| code lines added | 308 | 348 |
| ratio | 1 in 5.1 | 1 in 5.8 |
| `kubectl` sites removed | — | 1 |
| `.Logs(` calls added | — | 2 |

Some of the new comments are on `resolve` and `splitRef` and are fine — the port rule is worth stating. But across three pushes the narration is where it started, and eight suites still shell out to `kubectl` while `Logs` has two callers.

What I would find easier is a reply rather than a push. Both of these are decisions rather than defects, and either answer is fine:

- **Comments** — if you think the ratio is right for this change, say so and I will drop it. What I am reacting to is that it is roughly double your recent e2e work on the change least in need of explanation, not that any single comment is wrong.
- **`Logs`** — doing the conversion here, or opening the follow-up with the in-scope count in the body, both work. What I want to avoid is it landing with two callers and no record, which is the same shape as `fixture.ReceiverURL` sitting unused while nine sites spelled the literal inline.

The rest of the PR has been ready for a while: three image declarations collapse to one, the import-cycle reasoning for `internal/image` holds, and `TestAllMatchesTheLoaderList` is the check that stops a drifting env var loading one image while the suite runs another. I would like to merge it — these two are the only things between here and that.

vyncint:
The `Env` fix — agreed, and nothing further from me on that or the copilot thread.

**Comments.** My answer is that the level is right for this change, and here is the arithmetic, because I think the metric is measuring something else. Of the 60 added comment lines, **39 are Apache headers**: the diff adds three files and the header is thirteen lines each, which `make license-check` requires. That leaves 21 prose, and **8 of those moved unchanged** with `Receiver` out of `harness.go` into its own file. So 13 lines are newly written against 348 of code, four of them the `resolve` and `splitRef` rules you have already called worth stating. The ratio doubles against the recent e2e work because that work edited existing files and this one creates three — it is counting new files, not narration. I did cut seven lines on the first ask, and I will cut more if you still want it after that split; I only did not want to trade away the two decision-recording comments to move a number that is mostly licence text.

**`Logs`.** The count is already in the PR body under "Receiver log adoption — split out" — it went in shortly before your last comment, so it likely crossed. It is 17 sites rather than 18: the elasticsearch four is three, all `kubectl exec` on the curl pod. Seven convert, being every `kubectl logs`, all receiver reads selecting on `ReceiverName`. The other ten cannot — six `kubectl exec`, four `kubectl scale` on the receiver deployment `volumedrain` drains against. It is split rather than skipped because five of the seven pass `--tail 30` while `volumedrain`'s two read unbounded, so a shared helper has to decide that rather than inherit it, and bounding a read changes what a `Contains` check can see. I can open the follow-up issue too if you want the record off this PR.

csatib02 (approved):

PR #2323, opened 2026-08-20
annotations for logging-operator deployment in charts/logging-operator

Resolt:
deploymentAnnotations added to values.yaml and templated into logging-operator deployment.
needed when using tools such as keel.

### DIFF
--- a/charts/logging-operator/values.yaml
+++ b/charts/logging-operator/values.yaml
@@ -37,6 +37,9 @@ namespaceOverride: ""
 # -- Define annotations for logging-operator pods.
 annotations: {}
 
+# -- Define annotations for logging-operator deployment
+deploymentAnnotations: {}
+
 # -- Deploy CRDs used by Logging operator.
 createCustomResource: false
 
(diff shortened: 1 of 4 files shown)


Resolt:
The failing diff test needs to fail in this case, as documentation (for the helm chart) has to change for the feature to take effect.

charts/logging-operator/values.yaml
```diff
@@ -293,33 +295,33 @@ logging:
...
+    #   containerOverrides:
+    #   image:
+    # -- configure systemdTailers of HostTailer
+    # systemdTailers:
+    # - name: system-sample
+    #   disabled: false
+    #   systemdFilter: kubelet.service
+    #   maxEntries: 20
+    #   containerOverrides:
+    #   image:
```
csatib02: Uncommenting this breaks the chart, please revert this hunk.

csatib02 (commented):
Thanks, please run `make generate` and commit the changes.

charts/logging-operator/values.yaml
Resolt: I've opted for fixing the nested commenting instead in new force push - hope this will suffice

Resolt:
> Thanks, please run `make generate` and commit the changes.

make generate                       
Makefile:280: *** target pattern contains no '%'.  Stop.

I'm unsure why this happens. I've looked around and can't initially figure out what I'm missing in local config.

csatib02 (approved):

PR #2327, opened 2026-08-25
Add enableServiceLinks flag

### LINKED ISSUE #2326: Support enableServiceLinks for Fluent Bit / Fluentd pods
FluentbitSpec, FluentdSpec, HostTailerSpec or  EventTailerSpec don't expose enableServiceLinks. In namespaces with many Services, the default value (true) causes every Fluent Bit/Fluentd container to get {SVC}_SERVICE_HOST/{SVC}_PORT_* env vars injected for every Service in the namespace which is bloating the environment and leaking internal service topology.

Add an enableServiceLinks field (bool) to FluentbitSpec, FluentdSpec, HostTailerSpec, EventTailerSpec, set on the generated DaemonSet/StatefulSet pod template.

A generic pod-spec override object (like SyslogNGSpec.statefulSet) would also do this, but a single dedicated field is simpler and matches the existing model.

**Additional context**
Since LoggingSpec.fluentbit/LoggingSpec.fluentd embed these same specs, this also fixes it for the legacy Logging resource with no extra work.

nak0f:
Fixes https://github.com/kube-logging/logging-operator/issues/2326


csatib02 (commented):
Thanks @nak0f,

Can you wire this up for syslog-ng too?

csatib02 (approved):

PR #2328, opened 2026-08-25
refactor(e2e): move the detached suites onto the harness

vyncint:
The two detached suites, which finishes the `CheckExcess*` conversion #2324 started.

| suite | lines |
| --- | --- |
| `fluentd-aggregator-detached` | 311 → 145 |
| `syslog-ng-aggregator-detached` | 356 → 167 |

### The create inside the poll

Both suites created the excess config *inside* the `Eventually`, once the list showed only one, and tolerated `AlreadyExists` to survive the retries. A `Condition` reads, so it cannot do that — but it does not need to. The create only had to happen after the first config attached, which is a sequence rather than a loop:

```go
env.WaitFor(..., wait.LoggingUsesFluentd(loggingName, attachedFd), wait.AttachedFluentd(ns, attachedFd, loggingName))

env.Create(detachedFluentd(excessFd, 0))
env.WaitFor(wait.ExcessFluentd(ns, excessFd))
```

The `AlreadyExists` tolerance goes with the retry that required it.

### Conditions

Six, over two predicates. `attached` is no problems, names that Logging, active; `excess` is the inverse. `LoggingUsesFluentd` and `LoggingUsesSyslogNG` cover the other direction, which the Logging writes before the config's own status settles.

That leaves `CheckFluentdStatus` and its three siblings with no caller outside their own tests, and `refresh` existed only to serve them, so all five go. Their coverage of the per-direction `ptr.Deref` default moves onto the predicates: an unset `Active` still has to leave both unsatisfied, or an unreconciled config reads as settled.

The commits are ordered so the removal lands after the callers move; each builds, vets and tests on its own.

### Verification

`golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`.

Against real clusters, with `fluentbit-multitenant` as an untouched control: `fluentd-aggregator-detached` 99s, `syslog-ng-aggregator-detached` 223s, control 92s. Both dumps show the operator attaching one config and rejecting the other, so the excess conditions are observing a real verdict rather than passing early — an unreconciled config has no problems recorded, so `excess` cannot be satisfied before the operator writes it.

The predicates were checked against mutations rather than assumed to bite: flipping either `ptr.Deref` default, or dropping the `loggingName` comparison, each fails a test.


e2e/syslog-ng-aggregator-detached/syslog_ng_aggregator_detached_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2021 Cisco Systems, Inc. and/or its affiliates
```
csatib02, line 1: No need to change the header of a pre-existing file.

csatib02:
`detachedSyslogNG` and `detachedFluentd` are in the right place, for the record, since suite-local CR constructors are the thing we deliberately stopped putting in `fixture`.

Both have two callers inside their own file — the attached config and the excess one — so they are deduplicating within a suite rather than across suites. That is the case the #2308 rule allows: the literal is the suite's own specification, and a constructor next to it cannot let a default reach a suite that did not ask for it. Promoting either into `fixture` would be re-making the mistake the `WithFluentbit` and `WithFluentd` builders were deleted for.

The asymmetry between them is right too. `detachedFluentd` takes `workers` because that suite genuinely varies it, 2 for the attached config and 0 for the excess one, while `detachedSyslogNG` takes only a name because nothing there diverges. Worth leaving as is rather than making the signatures match — a parameter that exists for symmetry is the first step back toward defaulted builders.

The rest of this reads well. Replacing the create-inside-the-`Eventually` with a sequence is better than the `AlreadyExists` tolerance it removes: the create only ever needed the first config to have attached, and tolerating a duplicate was working around the retry rather than the ordering. Opposite `ptr.Deref` defaults on `attached` and `excess` are the detail most likely to be got wrong here, and having an unset `Active` leave both unsatisfied is what stops an unreconciled config reading as settled.

csatib02 (approved):

PR #2329, opened 2026-08-25
fix(model): mark defaultFlow outputs active

sebastiangaiser:
The validation reconciler reset every fluentd ClusterOutput to Active=false and only flipped it back for Spec.ErrorOutputRef or refs coming from a Flow/ClusterFlow. Outputs referenced solely by Logging.spec.defaultFlow.globalOutputRefs stayed inactive even though FlowForDefaultFlow renders them into the fluentd config, so both the status field and the resource state metrics reported an in-use output as unused.

Add a regression test covering a ClusterOutput reachable only through defaultFlow.


csatib02 (commented):
Verified the fix and the tests locally. LGTM.

csatib02 (approved):

PR #2330, opened 2026-08-26
refactor(e2e): move fluentd-aggregator onto the harness

vyncint:
`fluentd-aggregator`, the largest suite: 805 lines to 271, four tests and four clusters.

The specs share a builder per layer, with `Workers`, the read-only root filesystem and the config-check strategy passed in, since those are what the individual tests are about. Nothing is defaulted that a test varies.

### One check was passing for the wrong reason

The first wait in `TestFluentdAggregator_ConfigChecks` read `ProblemsCount` off the local `logging` value, which was never re-read after `Create`:

```go
return logging.Status.ProblemsCount == 0
```

That is the Go zero value rather than the cluster's answer, so the condition could not fail. `wait.LoggingHealthy` reads the object back. It still passes, so nothing was hiding behind it, but it was asserting nothing.

### Conditions

- **`PodFinished` / `ConfigCheck`.** The counterpart of `PodRunning`: a config check reports by exiting rather than by staying up, and both terminal phases count.
- **`LoggingHealthy`, `LoggingProblem`, `LoggingProblemCleared`.** The status the operator writes back. The matcher is a function because a failing check names the checksum in its message, so the problem is only recognisable by shape.

`Cleared` is not the negation of `Healthy`, and the test says so: an unrelated problem leaves the check's own one cleared while the Logging is still unhealthy.

`AnyPodShouldBeFinished` has no caller left and goes in its own commit, after the suite that used it moves.

### Verification

`golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`, and all three commits build, vet and test on their own.

Against real clusters, with `fluentbit-multitenant` as an untouched control: `MultiWorker` 101s, `ConfigChecks` 87s, `DryRun` 51s, `StartTimeout` 83s, package 188s, control 93s.

Neither new wait can pass vacuously. `LoggingProblem` needs a matching problem present, and the ordering forces it to appear and then clear, so the break-and-fix sequence is doing the work. The predicates were checked against mutations rather than assumed to bite: accepting a running pod as finished, and collapsing `Cleared` into `Healthy`, each fail a test.

### DIFF
--- a/e2e/fluentd-aggregator/fluentd_aggregator_test.go
+++ b/e2e/fluentd-aggregator/fluentd_aggregator_test.go
@@ -15,33 +15,15 @@
 package fluentd_aggregator
 
 import (
-	"context"
-	"fmt"
-	"os"
-	"os/exec"
-	"path/filepath"
 	"regexp"
-	"slices"
-	"strings"
 	"testing"
-	"time"
 
-	"github.com/cisco-open/operator-tools/pkg/types"
-	"github.com/cisco-open/operator-tools/pkg/utils"
 	"github.com/stretchr/testify/require"
-	appsv1 "k8s.io/api/apps/v1"
-	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/cluster"
 
-	"github.com/kube-logging/logging-operator/e2e/common"
-	"github.com/kube-logging/logging-operator/e2e/common/setup"
 	"github.com/kube-logging/logging-operator/e2e/internal/harness"
 	"github.com/kube-logging/logging-operator/e2e/internal/image"
 	"github.com/kube-logging/logging-operator/e2e/internal/wait"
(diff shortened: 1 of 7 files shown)


csatib02:
Approving in substance — the suite goes 805 lines to 271 without a default absorbing a divergence, and `TimeoutSeconds: 30` staying visible at the test that needs it is the part I checked hardest. Finding the wait that read `ProblemsCount` off a never-refreshed local value is the fourth cannot-fail assertion turned up in this module, and the first found without someone asking.

**`PodRunning` and `PodFinished` differ only in which phases count.** Everything else — list, loop, return on match — is duplicated. Worth one shared form:

```go
func podInPhase(name string, phases []corev1.PodPhase, opts ...client.ListOption) Condition {
	// ... list, then:
	if slices.Contains(phases, pod.Status.Phase) {
		return true, nil
	}
}

func PodRunning(name string, opts ...client.ListOption) Condition {
	return podInPhase(name, []corev1.PodPhase{corev1.PodRunning}, opts...)
}

func PodFinished(name string, opts ...client.ListOption) Condition {
	return podInPhase(name, []corev1.PodPhase{corev1.PodSucceeded, corev1.PodFailed}, opts...)
}
```

`slices` is already imported in that file. Both exported names stay, so no call site moves, and `podInPhase` stays unexported.

**`requireCleanConfigCheck` has two callers and is a whole test body rather than a helper.** Both call sites are one lines, and the two tests differ only in cluster name and config-check strategy. That is a table:

```go
for _, c := range []struct {
	name, cluster string
	check         v1beta1.ConfigCheck
}{
	{"dry run", "fluentd-configcheck-dry-run-readonly-rootfs", v1beta1.ConfigCheck{Strategy: v1beta1.ConfigCheckStrategyDryRun, Labels: configCheckLabels}},
	{"start timeout", "fluentd-configcheck-start-timeout-readonly-rootfs", v1beta1.ConfigCheck{Strategy: v1beta1.ConfigCheckStrategyTimeout, TimeoutSeconds: 30, Labels: configCheckLabels}},
}
```

The five parameters become columns, and the two varying fields sit next to each other where the difference between the tests is the thing being read. Cluster count does not change, so nothing about the concurrency tuning moves.

**`var logging v1beta1.Logging` in that function shadows the package-level `logging` constructor.** It compiles because `logging(loggingName, ns, spec)` is called earlier in the same body, but after that line the constructor is unreachable by name. Worth renaming the variable — `deployed` or `current` — since the two are one letter apart in a file where both are used.

The rest reads well: `Cleared` not being the negation of `Healthy`, with a case saying why, is the kind of distinction that is cheap to lose. And removing `AnyPodShouldBeFinished` in its own commit after the last caller moves keeps every commit green on its own, same as #2328.

csatib02 (approved):

Issue #2325, opened 2026-08-24
E2E: the kubectl shell-outs that Receiver.Logs cannot replace


vyncint:
Following up on the `Receiver.Logs` adoption raised in #2322, with the counts measured rather than estimated, so the remainder is a recorded decision instead of something the next reader works out.

**Is your feature request related to a problem? Please describe.**

`internal/harness.Receiver` owns reading the test receiver, but suites still shell out to `kubectl` directly. Measured on `master` @ `0cd9ec06`, there are **17** `exec.Command("kubectl", …)` sites across eight suites: 7 `logs`, 6 `exec`, 4 `scale`.

The important part is what happens to them, which is not uniform.

**Every `kubectl logs` site is a receiver read, and every one is in a suite that has not migrated yet.** Migration replaces it with `env.Receiver.MustReceive(tag)` as a side effect — #2324 removes two of the seven that way, without touching the receiver API. So this is not a separate adoption task; it is a consequence of finishing the migration, and adding a package-level `ReceiverLogs` for the interim would create a helper whose only purpose is to be deleted again.

After #2324 lands, the position is:

| suite | `logs` | `exec` | `scale` |
| --- | --- | --- | --- |
| `volumedrain` | 2 | 2 | 4 |
| `elasticsearch-multiversion` | 0 | 3 | 0 |
| `fluentd-aggregator` | 1 | 0 | 0 |
| `fluentd-aggregator-detached` | 1 | 0 | 0 |
| `syslog-ng-aggregator-detached` | 1 | 0 | 0 |
| `logging_metrics_monitoring` | 0 | 1 | 0 |
| **total** | **5** | **6** | **4** |

All six are unmigrated, so the 5 `logs` go with their suites.

**What is left over, and is the actual subject of this issue**

The **10** that `Receiver.Logs` cannot replace, and which migration does not remove:

- **6 `kubectl exec`** — coverage collection (`kill -USR1`, `tar -cf`) in `logging_metrics_monitoring` and `volumedrain`, and three `curl` calls against Elasticsearch through a pod in `elasticsearch-multiversion`.
- **4 `kubectl scale`** — `volumedrain` scaling the test-receiver deployment down and back up, which is the thing that suite exists to drain against.

These are the seam #2291 named `internal/kube` for, and they need a different answer from the receiver: an exec helper and a scale helper on the cluster, or a decision to keep them as shell-outs. Worth settling once the migration is done rather than inventing the package early with two callers.

[...]

PR #2331, opened 2026-08-26
refactor(e2e): move volumedrain onto the harness

vyncint:
`volumedrain`: 500 lines to 226, and eight `kubectl` invocations to one.

That last number is the part worth reading, since it is the count from #2325:

| | before | after |
| --- | --- | --- |
| `kubectl scale` on the receiver | 4 | 0 |
| `kubectl logs` on the receiver | 2 | 0 |
| `kubectl exec` into the aggregator | 2 | 1 |

### `Receiver.Scale`

Taking the receiver away is how a drain test makes the aggregator buffer instead of deliver, and the harness already owns the receiver — its name, its URL, its logs. Scaling joins them, so the suite says what it wants rather than spelling an invocation. That is four of the ten calls #2325 listed as needing a different answer.

The remaining one lists a path inside the aggregator's container. It is the exec seam that issue tracks, and it stays a shell-out here rather than growing a helper with one caller.

### Conditions

`JobActive` is the drainer the operator starts when a replica goes away. Presence is not enough: the Job exists before it runs, so waiting on presence alone is satisfied while nothing is draining.

`JobGone`, `PodGone` and `PVCGone` share a `gone` predicate that reads `NotFound` as the answer rather than as an error, leaving anything else to poll.

Those retire `ResourceShouldBeAbsent`, `ResourceShouldBePresent`, `PodShouldBeRunning` and `common.Resource`, which had no other callers. `common/resources.go` goes with them. The removal is its own commit, after the suite that used them moves, so every commit is green on its own.

### Both tests

They share the drain arc and differ in what they assert at the end: one keeps the claim and reads the operator's verdict off its label, the other has it deleted with the replica. The arc is a helper rather than a table because the ending is the substance, not a column.

### Stated rather than silent

The buffer check polled at 10s in the delete-volume test and now polls at the shared 3s. Polling more often cannot change a verdict.

### Verification

`golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`, and all four commits build, vet and test on their own.

Against real clusters, with `fluentbit-multitenant` as an untouched control: `TestVolumeDrain_Downscale` 146s, `TestVolumeDrain_Downscale_DeleteVolume` 125s, control 81s. Both dumps show the drainer running against filled buffers, so the arc is exercised rather than skipped — `JobActive` cannot be satisfied by a Job that has not started, and `PVCGone` needs the claim actually collected.


csatib02:
The timing collapse needs a note, and one of the waits may be unsafe.

`volumedrain` had nine distinct durations across its waits, including a deliberate 1s poll on the drainer Job at the old `:453`. Everything now runs at the shared 3s.

The drainer Job is the shortest-lived object here, and `JobActive` tightens the check from presence to `Status.Active > 0`, which makes it more sensitive to the interval rather than less. If the Job finishes inside 3s, a poll can land before it starts and again after it ends, so `Active > 0` is never observed — the wait then burns its budget, or `JobGone` passes for a Job that was never seen draining.

Either keep a shorter interval on that pair, or say what the drainer's measured lifetime is and why 3s clears it.

The other half is the budgets: several 30s waits now get up to the shared 5m. On a drain test "gone within 30s" and "gone within 5m" are different assertions, and the second will not catch a drain-speed regression.

Both are the kind of divergence your earlier batches called out explicitly, so worth the same treatment here rather than leaving it in the diff.

csatib02 (approved):

PR #2332, opened 2026-08-26
refactor(e2e): move elasticsearch-multiversion onto the harness

vyncint:
`elasticsearch-multiversion`: 815 lines to 528, and `internal/wait` down to Conditions alone.

### `wait.Deployment`

Ready alone can hold while the Available condition still reports the old rollout, so all three have to agree. This is the Condition form of `DeploymentAvailable`, without the `testing.T` and `context` parameters and without logging from inside the predicate.

One tightening over the helper it replaces, stated rather than left to be noticed: that one bailed when `Spec.Replicas` was nil, this one reads it as zero and refuses. A Deployment scaled to nothing cannot satisfy a readiness wait on the strength of a stale Available condition.

### The three versions are a table

The Services and Deployments were 330 lines that differed in 16, all of them the name and the version string. Container spec, probes, resources, env and ports were identical, so they become a `{name, version}` table and two constructors.

The Outputs stay as they are. ES7 sets `LogstashPrefix` and `IndexName` where 8 and 9 set `DataStreamName`, so they differ in shape rather than in values, and a table over them would need a union of fields.

### The curl pod is a type

Three call sites spelled out the same exec, the same flags, the same `?h=count` and the same `TrimSpace`. One `hasDocuments(host, index)` puts the invocation and the parsing in one place. Suite-local, since no other suite has a curl pod.

### The readiness wait stays suite-local

`env.WaitWithin` would take the divided budget. It would not take the restart count logged on each failed poll, which a read-only Condition cannot produce, so the wait stays where it is and drives `wait.Deployment` instead of the old helper.

That is the first escape hatch the harness has needed across the suites, so it is recorded on `WaitWithin` itself rather than left for the next suite to rediscover.

### Verification

`golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`, and all three commits build, vet, test and lint on their own.

The table was checked against the literals it replaces rather than by eye: a temporary `cmp.Diff` of the generated Service and Deployment against the original es7 blocks, with 8 and 9 derived from es7 by name, version and image alone. The three `_cat/count` URLs were compared byte for byte with the originals.

Against real clusters, with `fluentbit-multitenant` as an untouched control: `TestElasticsearch_MultiVersion` 363s, control 171s.


csatib02:
Four things, in order of how much they change.

**The Service and Deployment triplets are near-identical.** The three Deployment blocks are about 86 lines each and differ in 20 lines, all of them the name and the version string:

```
elasticsearch7 / 7.17.16   ->   elasticsearch8 / 8.12.0
```

Container spec, probes, resources, env, ports and volumes are the same in all three. That is roughly 340 lines where a three-row table over `{name, version}` would do, and it is most of why this suite lands at 743 rather than near the ~150 the restructure issue estimated for it.

The "suite's own specification" argument does hold for the Outputs, and I would leave those alone: ES7 uses `LogstashPrefix: "fluentd-es7"` while 8 and 9 use `DataStreamName: "logs-fluentd-es8"`. Those differ in *shape*, so a table needs a union of fields and stops being readable. The judgement is worth making per triplet rather than per suite — the Deployments collapse, the Outputs do not.

**`harness.Builder.WithScheme` has no caller.** Nothing in `e2e/` calls it; `schemeBuilders` is read only by `WithScheme` itself and passed to `buildScheme`. Every suite is covered by `defaultSchemeBuilders`. This is the last suite to migrate, so the evidence is complete in a way it was not for the earlier ones — if none of the thirteen needed extra builders, none will. Worth dropping the method, the field, the `extra` parameter and the subtests that exercise it; it is a handful of lines to add back the day a suite needs a CRD outside the default set, with a caller attached.

**The curl pod wants to be a type.** Three call sites spell out the same exec, the same `-s`, the same `?h=count` and the same `TrimSpace`:

```go
cmd := common.CmdEnv(exec.Command("kubectl", "exec", curlPod.Name, "-n", ns, "--",
    "curl", "-s", "http://elasticsearchN.logging.svc:9200/_cat/count/<index>?h=count"), env.Cluster)
```

A suite-local struct with `get(url)` and `docCount(host, index)` puts the invocation and the parsing in one place and lets each wait read as `c.docCount("elasticsearch7", "fluentd-es7-*")`. Suite-local rather than on the harness, since this is the only suite with a curl pod — on the harness it would be the one-caller case again.

**The readiness wait staying suite-local is worth recording as a design signal.** Your reason is right: `WaitWithin` takes the divided budget but cannot carry the per-poll restart-count diagnostic, because a read-only `Condition` cannot log. That makes it the first escape hatch the harness has actually needed across thirteen suites. One is a signal rather than a problem, but it is the kind that should be written down, since the next suite wanting per-poll diagnostics hits the same wall and the answer should not be rediscovered.

Separately, `wait.Deployment` is a faithful port with one tightening that is not in the description: the old helper bailed when `Spec.Replicas` was nil, the new one takes `ptr.Deref(..., 0)` and rejects `want == 0`, so an explicit `replicas: 0` is also caught. Correct for a readiness wait, worth a line so it is not read as accidental.

csatib02:
I'm going to close this, as #2334 contains both and I'm fine with one pr for all!

PR #2334, opened 2026-08-27
refactor(e2e): move logging_metrics_monitoring onto the harness

vyncint:
The last suite, and `common.WithCluster` with it. Stacked on #2332 — review that one first; this diff shows three extra commits until it merges.

### `logging_metrics_monitoring`

420 lines to 324. The helm install stays exactly as it was: `watch-selector` already installs a chart from a remote repo while running on the harness, so this only changes where it reads the kubeconfig from. What goes is the usual boilerplate — the cluster callback, the teardown closures, the temp directory and the scheme block.

The prometheus-operator types are the exception and stay explicit, as `WithScheme(v1.AddToScheme)`. This is the only suite that installs their CRDs, which is what that builder method exists for.

The two `Logging` literals become named constructors. They differ in the aggregator and share Fluent Bit, which is the point of the second half: its metrics have to survive the aggregator being replaced.

`common.WaitForPodReady` is satisfied by Running alone despite its name, so `wait.Pod` is the same assertion and says so. The three metric polls become one `mustServe`; reading them means curling from inside a pod, which a read-only Condition cannot do, so the poll stays suite-local.

### `common.WithCluster`

Every suite is on the harness now, so nothing calls it. The const that bounded its stop wait goes with it, and `WaitForPodReady` becomes unexported — `SetupCurlPod` is its only caller left, and a suite reaching for it instead of `internal/wait` is what the harness exists to prevent.

### Verification

`golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` reports 0 issues in `e2e/`, and both commits build, vet, test and lint on their own.

The rewrite was checked against the literals it replaces rather than by eye: a temporary `cmp.Diff` of both `Logging` specs against the originals, and an assertion that all six derived ServiceMonitor names and both pod names are unchanged.

Against real clusters, with `fluentbit-multitenant` as an untouched control: `TestLoggingMetrics_Monitoring` 329s, control 81s.


e2e/internal/wait/condition_test.go
```diff
@@ -151,3 +152,64 @@ func TestJobStartedCountsTerminalStatesToo(t *testing.T) {
...
+// unready, and the Available condition can still be True from before it was
+// scaled down.
+func TestDeploymentWantingNoReplicasIsNeverReady(t *testing.T) {
+	zero := int32(0)
+	for _, c := range []struct {
+		name     string
+		replicas *int32
+	}{
+		{"replicas unset", nil},
+		{"explicitly zero", &zero},
```
csatib02, line 184: you can use the `new()` built-in.

csatib02:
The failing `Test` job is `make check-diff` on `e2e/go.mod`:

```
-	github.com/go-logr/zapr v1.3.0 // indirect
-	go.uber.org/zap v1.28.0 // indirect
```

Deleting `common.WithCluster` took the last `ctrl.SetLogger(zap.New(...))` with it, so both indirects are unused now. `cd e2e && go mod tidy` and commit `go.mod` with `go.sum`.

Two things I checked rather than took on trust, since both could have quietly weakened an assertion.

**`WaitForPodReady` really is Running alone.** The name says otherwise, but the readiness loop cannot change the result:

```go
isReady := updatedPod.Status.Phase == corev1.PodRunning
for _, cond := range updatedPod.Status.Conditions {
	if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
		return true, nil
	}
}
return isReady, nil
```

The only case the loop adds is Running false with PodReady true, which does not occur. So `wait.Pod` is the same assertion under a name that matches it. It also polls on a missing pod through `client.IgnoreNotFound` rather than erroring, which is what a wait should do.

**`WithScheme` earns its keep here.** I asked for it to be deleted on the elasticsearch PR, on the grounds that no suite had needed it across the migration. That was wrong: I read that PR as the last one, and this is the last one. `WithScheme(v1.AddToScheme)` for the prometheus-operator CRDs is the caller the method was waiting for, and one caller in thirteen suites is the right amount of evidence for keeping it.

The rest reads well. Checking the two `Logging` constructors with a temporary `cmp.Diff` against the literals they replace, plus pinning the six derived ServiceMonitor names and both pod names, is the right way to do a rewrite this size — the equivalence is asserted rather than reviewed by eye.

Worth saying plainly what this one is: `common.WithCluster` going is the change #2291 called the single highest-leverage one in the whole restructure, and this retires it along with the last suite.

csatib02 (approved):

